PR #18, opened 2015-03-30
svgo has been updated to 0.5.1

GreLI:
Also, a new command-line option to set precision has been added. But I don't know how to correspondingly update svgomg to it.


jakearchibald:
Brilliant, I'll look at that option and how to integrate it

PR #34, opened 2015-10-21
Fix npm 3 support

nolanlawson:
npm 3 has flat dependencies, so the regeneratore/runtime dependency
can't be deeply nested anymore.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "merge-stream": "^0.1.6",
     "pako": "^0.2.7",
     "prismjs": "git+https://github.com/LeaVerou/prism.git#gh-pages",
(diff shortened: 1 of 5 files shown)


jakearchibald:
Whoa, I managed to miss this. Will take a closer look at this rather than just simply merging. Last time I flattened the regenerator include things started failing in a really odd way (would fail during fresh installs only).

DanReyLop:
Since npm3 is now the norm, I would say this is a high priority issue. Could you please take a look at it?
I've done a fresh install to @nolanlawson branch and everything works fine. I also applied the patch in an existing copy of the repo and works fine too.

PR #39, opened 2015-12-07
added charset

J0WI:
this may fix a charset issue when opening non utf-8 SVGs


jakearchibald:
I can't see how this would fix anything - the html is already served as utf-8. Can you give me a failing SVG?

PR #58, opened 2016-08-30
smaller icon.png

tomByrer:
before: 3.94 KB
after: 1.8 KB
Thanks to ImageAlpha & ImageOptim, users will able to download svgomg 0.01% faster.


jakearchibald:
Haha, brilliant, thank you!

PR #59, opened 2016-09-04
Enhance the ability to navigate by keyboard

shadeed:
This PR is about adding focus styles for the following elements:
- Main menu button
- Material tabs
- The toggle / switch list

Currently, it's hard to use the app without a mouse. My mouse was not working and I noticed that while trying to use it with the keyboard.

### DIFF
--- a/src/css/components/_material-switch.scss
+++ b/src/css/components/_material-switch.scss
@@ -44,10 +44,12 @@
     left: 0;
     width: 20px;
     height: 20px;
-    border-radius: 10px;
+    border-radius: 20px;
     background: #fafafa;
-    box-shadow: 0 1px 3px rgba(#000, 0.5);
+    box-shadow: 0 1px 3px rgba(0,0,0,0.5), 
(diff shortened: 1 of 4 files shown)


shadeed:
When navigating via keyboard from the Main menu button, it is a bit confusing that the focus won't move to the second tab "Code" when hitting tab.

svgomg

That's because we are tabbing through a group of radio buttons. Do you think that we can replace them with regular links?

jakearchibald:
Thanks for this! Kinda ashamed it wasn't so great. Will do a review & merge.

jakearchibald:
"Kinda ashamed it wasn't so great" - I should clarify that I mean I'm ashamed it wasn't great already. This PR looks great.

shadeed:
Hi @jakearchibald - No problem! I got what you meant 😃 

Glad to hear! this is the least I can do for a tool that I use almost everyday! <3

jakearchibald:
Merged! Will land on the site soon

PR #61, opened 2016-09-13
save settings when updating

### LINKED ISSUE #60: Remember config
Salut!

I use `SVGOMG` every single day, it is just great! But each time, I have to re-set my config. I don't thing it would be really hard to implement a **cookie system** that would **remember** our config.

Math

psimyn:
fixes #60

Potential issues:

- no opt-out / reset
- a better way to update range slider

### DIFF
--- a/src/js/page/main-controller.js
+++ b/src/js/page/main-controller.js
@@ -75,7 +75,7 @@ class MainController {
       if (document.queryCommandSupported && document.queryCommandSupported('copy')) {
         minorActionContainer.appendChild(this._copyButtonUi.container);
       }
-      
+
       actionContainer.appendChild(this._downloadButtonUi.container);
       document.querySelector('.output').appendChild(this._outputUi.container);
       this._container.appendChild(this._toastsUi.container);
@@ -146,6 +146,7 @@ class MainController {
   }
 
   _onSettingsChange() {
+    this._saveSettings();
     this._compressSvg();
   }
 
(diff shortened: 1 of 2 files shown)


src/js/page/ui/settings.js
```diff
@@ -67,14 +67,32 @@ class Settings extends (require('events').EventEmitter) {
...
+  setSettings(settings) {
+    this._globalInputs.forEach(function(inputEl) {
+      if (inputEl.type == 'checkbox') {
+        inputEl.checked = settings[inputEl.name];
+      }
+      else if (inputEl.type == 'range') {
+        inputEl.value = settings[inputEl.name];
+        var rangeChangeEvent = utils.isIe ? 'change' : 'input';
+        var event = new Event(rangeChangeEvent);
+        inputEl.dispatchEvent(event);
```
jakearchibald: Why dispatch an event here? I don't think setting `checked` triggers events.

psimyn: This is only for the range - just updating the input didn't update the material-slider elements so ui was out of sync

src/js/page/main-controller.js
```diff
@@ -155,6 +156,7 @@ class MainController {
     try {
       this._inputSvg = await svgo.load(event.data);
       this._inputFilename = event.filename;
+      await this._loadSettings();
```
jakearchibald: Any objections to doing this on DOM ready? Happy to make that change as I merge

src/js/page/ui/settings.js
jakearchibald: Ahhh gotcha, I'll work around this

src/js/page/main-controller.js
psimyn: :+1:

jakearchibald:
It's about time I started a contributors file. What name should I use there?

psimyn:
Simon Knox or psimyn, depending on format

PR #62, opened 2016-09-13
let drag and drop overlay appear multiple times

### LINKED ISSUE #57: Drag & drop overlay not appearing a second time
I really enjoy the drag and drop function but when it comes to the situation where you have two or more graphics you want to send through this brilliant machine i always have to reload the browser to be able to drag & drop again. Could you implement this functionality to work every time when clicking on the menu-icon, not only after the app loads?

psimyn:
fixes #57


jakearchibald:
Well spotted

jakearchibald:
Merged!

PR #75, opened 2016-11-02
Strip null terminator appended by Illustrator

### LINKED ISSUE #68: `Load failed: Text data outside of root node.` when pasting from Illustrator
A recent (ish? it's been a while) update causes `Load failed: Text data outside of root node` to be thrown when pasting directly from Illustrator (ie copying from an artboard). This error is _technically_ correct, since Illustrator prepends a comment when copy/pasting SVG: 

```<!-- Generator: Adobe Illustrator 20.1.0, SVG Export Plug-In  -->```

That said, it would be nice if instead of throwing an error and canceling the paste process you instead 

- a) Threw a warning that some text was stripped but continued with loading
- b) Threw an error but allowed the user to continue anyways (stripping the excess text)

### PR DESCRIPTION
**Pull request by alexzaworski**
Copy/pasting directly from an Illustrator artboard yields a null-terminated string, causing SVGO to throw `Text data outside of root node.`

Stripping the null terminator from the paste input's value fixes the issue and closes #68.

### CODE REVIEW

File: src/js/page/ui/main-menu.js [hunk 1]
```diff
@@ -63,7 +63,7 @@ class MainMenu extends (require('events').EventEmitter) {
   }
 
   _onTextInputChange(event) {
-    var val = this._pasteInput.value.trim();
+    var val = this._pasteInput.value.trim().replace(/\0$/, "");
```
alexjlockwood: FYI, according to this StackOverflow post, this will replace only one instance of a null terminator (not multiple): https://stackoverflow.com/a/22809513/844882

skycommand:
```suggestion
    var val = this._pasteInput.value.trim().replace(/\0+$/g, "");
```
I suggest adding `+` to catch sequential nulls and `/g` to make the search global.

PR #78, opened 2017-01-18
Add remove style element plugin

jessetane:
https://github.com/svg/svgo/blob/master/plugins/removeStyleElement.js

This plugin is disabled by default so should be a relatively inoffensive addition.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "regenerator": "^0.8.41",
     "regenerator-runtime": "^0.9.5",
     "run-sequence": "^1.0.1",
-    "svgo": "0.5.6",
+    "svgo": "0.7.1",
     "uglifyify": "^2.6.0",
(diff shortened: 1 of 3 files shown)


jakearchibald:
Thanks! Sorry for the delay getting to this, I'll get to this soon.

jessetane:
Thanks, no worries. I'm working with a designer who uses Illustrator which is generating a huge amount of style data we don't need.

4muzar:
@jakearchibald 
Hi! Merge it, please!=)  
This PR is also useful because of svgo upgrade! Newer version do work better than the old one.

jakearchibald:
I've got the next version in https://github.com/jakearchibald/svgomg/tree/next-version which includes updating SVGO. Just got some finishing touches to do.

alexjlockwood:
Don't think there is a need for this any longer. svgomg now includes this as an option.

PR #105, opened 2017-11-02
Fix possible typo in utils.js

alexjlockwood:
Unless you are referencing a global variable I'm not aware of... I think you meant to use `str` instead of `string`?

### DIFF
--- a/src/js/page/utils.js
+++ b/src/js/page/utils.js
@@ -25,7 +25,7 @@ const entityMap = {
   "/": '&#x2F;'
 };
 
-export function escapeHTML(string) {
+export function escapeHTML(str) {
   return String(str).replace(/[&<>"'\/]/g, s => entityMap[s]);
 }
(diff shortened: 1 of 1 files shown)


jakearchibald:
Well spotted!

alexjlockwood:
I was trying to learn how to use service workers and at the same time porting some of your code to Typescript. :)

jakearchibald:
I'm not against going TS. I've used it in some other projects so it isn't alien to me.

alexjlockwood:
I started out with Typescript mostly, so plain old JS can sometimes look pretty alien to me too. Although your code is much easier to understand than most. :)

BTW, I sent you a second pull request that deletes a now-unused-file (at least I think it is unused).

jakearchibald:
> Although your code is much easier to understand than most

No has ever said that. Not even me.

PR #106, opened 2017-11-02
Delete unused load-scripts.js file

alexjlockwood:
I think this file is no longer necessary now that you've migrated to using ES6 modules?


jakearchibald:
Correct. I guess we can bring it back if polyfills are needed in future.

PR #108, opened 2017-11-02
Import EventEmitter consistently across files

alexjlockwood:
No other file imports EventEmitter using the default import, so I changed it to match the other files. 

I'm not sure if there is any difference...


jakearchibald:
I'm going to be out at standards meetings for the next week, but I've added you as a collaborator if you want to continue to push to the branch.

jakearchibald:
Really appreciate you looking at this!

PR #113, opened 2017-11-24
update package-lock.json file

alexjlockwood:
When I run `npm install` a slightly modified `package-lock.json` file is regenerated.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3355,14 +3355,6 @@
             }
           }
         },
-        "string_decoder": {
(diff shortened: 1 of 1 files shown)


alexjlockwood:
Hmm... I think the gulpfile needs to be transpiled using babel? I'll try to fix it. :p

alexjlockwood:
I'll close this for now and will recreate once the builds are working again. We could also just increase the node version used by travis too probably.

PR #117, opened 2017-11-29
Next version

jakearchibald:
Don't merge just yet.

I just need to update the changelog (which also updates the service worker).

@alexjlockwood does this look ok? It seems like there's a bit of a regression in terms of the compression ratio. Am I missing something?


alexjlockwood:
Nice! Seems good, yeah. I haven't tested though... will checkout the branch when I get back home.

BTW, did you notice any issues with my svgo pull request that fixes "clone"? I noticed your implementation was different (and avoided JSON parsing entirely, probably for the better?). I can update my svgo pull request if you want.

jakearchibald:
I think using JSON was always a bit of a hack, and lost all kinds of prototype info. I didn't intend to ignore your PR, I just didn't have access to it on a plane 😄.

Just noticed that the workers are loading a bit too late, so I'll fix that before shipping.

Any testing is appreciated!

alexjlockwood:
Yeah, I agree about the JSON stuff feeling kinda hacky. I like yours better (just not sure if svgo can use Maps yet? Is that a relatively new feature in js?)

jakearchibald:
@alexjlockwood In that case I'm happy to maintain the clone function in SVGOMG. I should remove it from the prototype really, no need for it to be there.

PR #129, opened 2018-01-18
More visual feedback on saving/increase

pgiraud:
With this PR I tried to enhance the increase / decrease file size feedback to the user by:
 - showing the size of the original size,
 - adding +/- in front of percent of change depending on if it increases or decreases,
 - using green / red color.

screenshot from 2018-01-18 13-47-07


src/css/components/_results.scss
```diff
@@ -15,4 +15,15 @@
     min-width: 11.3em;
     border-radius: 2px;
   }
+
+  .saving {
+    color: green;
+  }
+  .increase {
```
odahcam: i think this guy could me named as `increasing` just to be more concise with `saving`, or even change `saving` to `saved` and `increase` to `increased`.

src/js/page/ui/results.js
```diff
@@ -19,30 +19,42 @@ export default class Results {
...
-      this._diffEl.textContent = ' - no change';
+      this._diffEl.textContent = 'no change';
+      this._diffEl.classList.remove('increase');
+      this._diffEl.classList.remove('saving');
     }
     else if (size > comparisonSize) {
-      this._diffEl.textContent = ' - ' + round(size / comparisonSize * 100 - 100, 2) + '% increase';
+      this._diffEl.textContent = '+' + round(size / comparisonSize * 100 - 100, 2) + '%';
+      this._diffEl.classList.add('increase');
+      this._diffEl.classList.remove('saving');
```
odahcam: you could always remove the color classes before the ifs chain, in that way you would save two lines of code 😄

pgiraud: :+1:

src/css/components/_results.scss
pgiraud: I decided to change to more neutral names "success / error".

odahcam: good point, success in reducing size, but error is not a really error. 🤔

jakearchibald:
Ohh this looks really interesting. Will review next week.

jakearchibald:
Love it! I'm going to change the class names a little as I agree with @odahcam.

Also, I retrospect I think "60% saving" is confusing, and should just be "40%". I'll make a little update for this too.

Massive thank you for this, and to @odahcam for reviewing.

jakearchibald:
Also, making a slight adjustment as this isn't really readable at mobile sizes.
screen shot 2018-02-05 at 13 24 46

jakearchibald:
Merged manually. Should be live shortly. Thanks again!

PR #135, opened 2018-04-23
add open file keybinding

patrickkettner:
every time I use the site I hit ctrl-o and it ends up opening the file in the browser rather than the app and i'd like for that to not happen


jakearchibald:
Merged manually https://github.com/jakearchibald/svgomg/commit/9ffe9e2b87850fa852b2323314c7a9628344e4aa - thanks!

PR #136, opened 2018-05-04
Change 'prettify code' to 'prettify markup'

alexjlockwood:
I have OCD... :)

But I guess this should be changed as well since 'code' was changed to 'markup' in a recent update.


jakearchibald:
Good catch

PR #138, opened 2018-06-14
Fix Promise.all() await keyword usage.

wanderview:


jakearchibald:
lol thanks. This is such a silly mistake.

Issue #101, opened 2017-07-17
No option to reset to default optimisation settings in UI


jonbeddoes-avast:
I don't know why this is the case specifically, but I always encounter problems using this tool when experimenting with the optimisation settings. Regardless of the state of my cookies or whether I am in Incognito mode, I can't seem to just "reset" this tool to the default optimisation toggle values and it always remembers the settings I had last time, even when I don't want this.

davidklebanoff:
Yes, a simple "Restore default settings" button somewhere would be nice.

ericcornelissen:
Seems this issue can be closed @jakearchibald, agreed @jonbeddoes-avast

jonbeddoes-avast:
@ericcornelissen I will have a look and see shortly 👍

PR #141, opened 2018-07-24
Button to reset configuration

ericcornelissen:
Adds a new button to the main menu that can be used to reset the configuration of SVGO. As requested in #101 (and [simple-icons/#771](https://github.com/simple-icons/simple-icons/issues/771)).

### Design considerations
- ~~There is currently no "Are you sure" popup when clicking the reset configuration button, might be an idea to add it~~ (added with dfb7392)
- The button is currently in the mainMenuUI, could also be placed in the settingsUI. I'm fine either way 😄 

### Clarification of the code
- [main-controller.js:277](https://github.com/jakearchibald/svgomg/pull/141/files#diff-a4743a35a5a5dd4f27da459b1fa5ebeaR277) was added because the reset button calls `_compressSvg`, but may be called when no SVG has been selected yet.
- I'm emitting an event in [main-menu.js:97](https://github.com/jakearchibald/svgomg/pull/141/files#diff-8a2874b84bbf96aff0c3afad995acddeR97) mainly because I couldn't think of a better approach

### Tested on Desktop
- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] IE/Edge

### Tested on Mobile
- [ ] Chrome
- [ ] Firefox


jakearchibald:
Ohh thanks! Will take a look.

jakearchibald:
> There is currently no "Are you sure" popup when clicking the reset configuration button, might be an idea to add it

I think the best way to do this is via a toast (https://github.com/jakearchibald/svgomg/blob/master/src/js/page/main-controller.js#L158) which says "Options reset" but has "Undo" & "Dismiss" buttons. I'll take a look at that.

ericcornelissen:
> I think the best way to do this is via a toast (https://github.com/jakearchibald/svgomg/blob/master/src/js/page/main-controller.js#L158) which says "Options reset" but has "Undo" & "Dismiss" buttons. I'll take a look at that.


I've already [used a toast](https://github.com/jakearchibald/svgomg/pull/141/files#diff-a4743a35a5a5dd4f27da459b1fa5ebeaR243), so I'm comfortable adding it for you 😄

jakearchibald:
I'm going to have a hack around with this. I think the setting is currently in the wrong place – it should sit with the settings it resets. There might be a better way to encapsulate the behaviour too. I'll have a play and get you to review if that's ok?

ericcornelissen:
Of course it is, have a go with it 😄

jakearchibald:
https://github.com/jakearchibald/svgomg/pull/142 - how's this?

ericcornelissen:
> #142 - how's this?

Looks good 👍

jakearchibald:
Done in #142. Thanks @ericcornelissen for helping with the design!

ericcornelissen:
Np, glad it is added now 🎉

PR #142, opened 2018-08-07
Adding reset button

jakearchibald:


src/js/page/ui/settings.js
```diff
@@ -17,6 +18,10 @@ export default class Settings extends EventEmitter {
         document.querySelectorAll('.settings .global input')
       );
 
+      this._resetRipple = new Ripple();
+      const resetBtn = document.querySelector('.setting-reset');
+      resetBtn.appendChild(this._resetRipple.container);
```
ericcornelissen: Personally I find it a bit weird that the reset button itself is not on `this` but its ripple effect is. Perhaps ripple can be triggered in the local event listener, or a separate even listener?

jakearchibald: Looks like I put stuff on this even if it isn't used outside of the constructor. That seems like a weird thing for past-me to do, but I'll keep doing it for consistency.

ericcornelissen: Haha, alright 👍

PR #146, opened 2018-11-21
Select only svg code without results text

ppsirius:
When you whant copy svg code and try select all text by mouse or shortcut we have copied results statistic text.


XhmikosR:
Not sure what this solves exactly but it doesn't allow one to copy the results anymore which is bad IMHO.

PR #158, opened 2019-08-23
Upgrade svgo and add new plugins

### LINKED ISSUE #130: update svgo to 1.0.4
please update the bundled svgo

https://github.com/svg/svgo/releases/tag/v1.0.4

marcofugaro:
This PR upgrades svgo's version to the latest and adds the new plugins since last version.

I had to upgrade also gulp-sass because it was giving me problems on the install.

Closes #130


src/js/svgo-worker/index.js
```diff
@@ -80,16 +84,21 @@ const pluginsData = {
   collapseGroups,
   convertPathData,
   convertTransform,
+  convertEllipseToCircle,
   removeEmptyAttrs,
   removeEmptyContainers,
   mergePaths,
   removeUnusedNS,
+  removeOffCanvasPaths,
```
jakearchibald: This seems to throw "__dirname is not defined" with the demo image, and leaves the app in an unusable state.

marcofugaro: I will look into it!

marcofugaro:
@jakearchibald as it turns out, `removeOffCanvasPaths` is the only plugin to [require the SVGO library itself](https://github.com/svg/svgo/blob/10a302f73dd81122c8363ae164ba5a6af1de46e2/plugins/removeOffCanvasPaths.js#L9), and SVGO isn't built for browsers, this is the reason it breaks.

Some people already encountered this issue, in [this thread](https://github.com/svg/svgo/issues/1050#issuecomment-478237278) about using SVGO in the browsers. SVGO would need to be transpiled through browserify to be used in the browsers, and that can't happen until https://github.com/svg/svgo/pull/1122 lands. Otherwise some refactoring in the `removeOffCanvasPaths` plugin needs to happen.

I commented out the plugin like the others.

jakearchibald:
This is great, thank you very much!

dr-js:
Maybe try the portable webpack output https://unpkg.com/libsvgo/webpack/SVGO.js from [dr-js/libsvgo](https://github.com/dr-js/libsvgo), a re-formatted ES6+ fork of svgo.

PR #175, opened 2019-12-03
Fix #124 - Timeout transitionend event

### LINKED ISSUE #124: Switching to code view, toggling the "show original" setting and then trying to return to image view fails
In Firefox 57.0.3 on Windows, switching to code view and then toggline the *show original* setting prevents the user from returning to image view again (it remains locked to code view) for a pasted image.

NotWoods:
Occasionally the image/markup tabs get stuck in Firefox and you can no longer switch between them. It seems that the `"transitionend"` event sometimes doesn't fire, and as a result the previous promise in the queue never resolves.

The quick fix is to add a timeout with `Promise.race`. Neither promise rejects, so it's sufficient here.

### DIFF
--- a/src/js/page/ui/view-toggler.js
+++ b/src/js/page/ui/view-toggler.js
@@ -1,9 +1,13 @@
 import { EventEmitter } from 'events';
 import { domReady } from '../utils';
 
+/**
+ * Tabs that toggle between showing the SVG image and XML markup.
+ */
 export default class ViewToggler extends EventEmitter {
   constructor() {
     super();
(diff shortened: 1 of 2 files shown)


XhmikosR:
@jakearchibald can you check this PR please? I keep hitting this on Firefox. Thanks!

jakearchibald (approved):

jakearchibald:
Thanks for fixing this, and sorry this repo doesn't get a lot of my attention. Thanks also for pinging me @XhmikosR!

PR #197, opened 2020-08-16
replace <symbol> with <svg> when pasting input

mustafa0x:
SVG `<symbol>`s are quite common, and the workflow for viewing them in SVGOMG is cumbersome. This should help.

### DIFF
--- a/src/js/page/ui/main-menu.js
+++ b/src/js/page/ui/main-menu.js
@@ -69,7 +69,13 @@ export default class MainMenu extends EventEmitter {
   }
 
   _onTextInputChange(event) {
-    const val = this._pasteInput.value.trim();
+    let val = this._pasteInput.value.trim();
+
+    // Replace <symbol> with <svg> so that the graphic is rendered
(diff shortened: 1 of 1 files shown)


jakearchibald:
Shouldn't this change go into SVGO rather than SVGOMG? Wouldn't this break with nested symbols?

mustafa0x:
> Shouldn't this change go into SVGO rather than SVGOMG?

I don't believe so, but maybe I'm missing something.

> Wouldn't this break with nested symbols?

Probably, and a set of heuristics to cover common use cases would probably be useful here too. E.g., if the user is copying a single `<path>`, a group, and so on. Not forcing the user to wrap valid SVG tags in `<svg xmlns="...">` is a boon.

`<symbol>` is the most common situation for me, as I store each icon as a `<symbol>`. I can't say whether it's common for others, but it may be.

jakearchibald:
I'm happy with SVGOMG being used for valid SVG for now. True, some partial elements can be independently rendered if they have a viewbox/width+height, but I don't think transforming them to `<svg>` is a good answer.

It's an interesting usecase to think about in future though.

PR #225, opened 2021-09-08
Show SVGO version in nav bottom

TrySound:
It can be not clear for user which svgo version is used under the hood.

image
image


jakearchibald (approved):

jakearchibald:
I tweaked it a little so the mid-link wrapping is prevented in CSS.

PR #229, opened 2021-09-08
Update navigation link to README.md

paulrobertlloyd:
Branch name was changed from `master` to `main`. The link currently redirects, but this avoids the need to do so.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -68,7 +68,7 @@
               </a>
             </li>
             <li>
-              <a href="https://github.com/jakearchibald/svgomg/blob/master/README.md" class="menu-item">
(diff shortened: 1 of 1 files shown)


paulrobertlloyd:
The Travis config also points to `master`, FYI: https://github.com/jakearchibald/svgomg/blob/44538bfe88541657b880d2ac4b394f4c7f9402cf/.travis.yml#L11

jakearchibald (approved):

jakearchibald:
Ohh good catch! The travis thing hasn't worked for a while, so I should probably just delete that file.

PR #231, opened 2021-09-09
Remove sass mixins for prefixes

TrySound:
Most prefixes are not necessary in latest browsers.
user-select, animation and keyframes are stil prefixed.

Default browserslist config is `> 0.5%, last 2 versions, Firefox ESR, not dead`


jakearchibald:
Do you think we need autoprefixer? There are so few cases where prefixes are needed now, they can just be done manually.

TrySound:
Dunno user-select, animation, keyframes, min-content, max-content are still prefixed in popular browsers

jakearchibald:
`user-select` needs a webkit prefix, but I don't think the others have needed a prefix for a long time.

- https://caniuse.com/mdn-css_properties_user-select
- https://caniuse.com/css-animation
- https://caniuse.com/mdn-css_at-rules_keyframes
- https://caniuse.com/mdn-css_properties_width_min-content
- https://caniuse.com/mdn-css_properties_width_max-content

TrySound:
Ok, makes sense since it's development tool

jakearchibald (approved):
LGTM! Should make the CSS significantly smaller.

src/css/_utils.scss
```diff
@@ -1,16 +1,7 @@
-@mixin placeholder {
-  &::-webkit-input-placeholder,
-  &::-moz-placeholder,
-  &:-ms-input-placeholder {
-    @content;
-  }
-}
-
 @mixin user-select($spec...) {
   -webkit-touch-callout: $spec;
```
XhmikosR, line 2: @TrySound Isn't this obsolete? It could probably be removed IIRC.

TrySound: 🤷‍♂️ Maybe

PR #232, opened 2021-09-09
Configure htmlmin's clean-css options

XhmikosR:
Should fix:

```
Source map error: Error: request failed with status 404
Resource URL: https://jakearchibald.github.io/svgomg/
Source Map URL: head.css.map
```

Another solution would be to disable sourcemaps for `head.css`.


TrySound:
We need to fix source maps not kill them. I almost fixed this locally btw

XhmikosR:
Closing then, I personally wouldn't bother with the critical CSS sourcemap, but if you fixed it, perfect :)

PR #233, opened 2021-09-09
Inline menu logo image.

XhmikosR:


src/index.html
```diff
@@ -32,7 +32,7 @@
         <div class="overlay"></div>
         <nav class="menu">
           <div class="app-title">
-            <img src="imgs/icon.svg" width="1" height="1" />
+            <svg viewBox="0 0 600 600" xmlns="http://www.w3.org/2000/svg"><title>SVGOMG</title><path fill="#0097A7" d="M0 1.995h600V600H0z"/><path fill="#00BCD4" d="M0 0h600v395.68H0z"/><path d="M269.224 530.33L519 395.485H269.224V530.33zM214.35 91.847H519v303.638H214.35V91.847z" opacity=".22"/><path fill="#FFF" d="M80 341.735h189.224V530.33H80z"/></svg>
```
jakearchibald, line 35:
```suggestion
            <svg viewBox="0 0 600 600"><path fill="#0097A7" d="M0 1.995h600V600H0z"/><path fill="#00BCD4" d="M0 0h600v395.68H0z"/><path d="M269.224 530.33L519 395.485H269.224V530.33zM214.35 91.847H519v303.638H214.35V91.847z" opacity=".22"/><path fill="#FFF" d="M80 341.735h189.224V530.33H80z"/></svg>
```

Don't need `xmlns` for inline SVG. Also, I removed the title since the title is already in the h1.

XhmikosR: Yeah I wasn't sure about these either, but I noticed you've kept the `title` in menu for example. a11y-wise it might not be right, but we can do an a11y pass later :)

jakearchibald (approved):

PR #235, opened 2021-09-09
Fix head.css source maps

TrySound:
Currently head.css source maps are not loaded because the its path is
changed when inlined into index.html which doesn't have head.css.map in
the same folder.

I suggest to move css to top level as the easiest way to fix this
and avoid paths manipulation.


XhmikosR:
An alternative solution is #232: I personally never bother with the critical CSS sourcemaps since it's also loaded later.

jakearchibald (approved):

TrySound:
@XhmikosR Removing comment doesn't fix source maps
image

PR #236, opened 2021-09-09
Replace http-server with sirv-cli

XhmikosR:
Much lighter: https://packagephobia.com/result?p=http-server%2Csirv-cli

BTW the `&` doesn't work on Windows native cmd. Would it be OK with you @jakearchibald to add `npm-run-all`?


jakearchibald (approved):

jakearchibald:
> Would it be OK with you @jakearchibald to add `npm-run-all`?

Yep, I'm happy with that

PR #239, opened 2021-09-09
Update patch and minor dependencies and flatten package-lock.json

XhmikosR:
I used npm 7.x although personally I still use npm 6.x which ships with the current Node.js LTS.

I kept prismjs back, but it should also be updated later since it has a high security issue:

```
prismjs  <=1.23.0
Severity: high
Regular Expression Denial of Service - https://npmjs.com/advisories/1638
Regular Expression Denial of Service - https://npmjs.com/advisories/1762
```


jakearchibald:
I'm not too worried about these kinds of issues, as they aren't the kind of things that can be exploited on a static site like this. But it's always good to update deps 😄

PR #241, opened 2021-09-09
Fix undefined variable

XhmikosR:
Not sure if that's the intention but `key` was unused and `id` was not defined.


jakearchibald:
Good catch! This is why I use typescript now 😄

XhmikosR:
BTW I will add GitHub Actions CI later; this would have been caught with
ESLint.

PR #243, opened 2021-09-10
Enforce LF

XhmikosR:
On Windows `autocrlf` is true by default, so this patch ensures we always use LF regardless.


XhmikosR:
@jakearchibald maybe you could disable Netlify's PR comments? It's extra noise :)

jakearchibald (approved):

PR #247, opened 2021-09-10
index.html: minor consistency changes

XhmikosR:


src/index.html
```diff
@@ -3,15 +3,15 @@
...
     <meta charset="utf-8">
     <title>SVGOMG - SVGO's Missing GUI</title>
-    <meta name="theme-color" content="#303F9F">
+    <meta name="theme-color" content="#303f9f">
     <meta name="viewport" content="width=device-width, minimum-scale=1.0">
     <link rel="preload" as="script" href="js/page.js">
     <link rel="preload" as="script" href="https://www.google-analytics.com/analytics.js">
     <link rel="preload" as="style" href="all.css">
-    <link rel="preload" as="style" href="https://fonts.googleapis.com/css?family=Roboto:400,700|Inconsolata">
+    <link rel="preload" as="style" href="https://fonts.googleapis.com/css?family=Roboto:400,700%7CInconsolata">
```
XhmikosR, line 11: @jakearchibald I think this is causing a mismatch with the one in line 189. But `|` is an invalid character. Should we change line 189 to use the escape sequence too?

jakearchibald: Ah yes. Sorry, I should have caught that.

XhmikosR: And probably the SW too.

PR #248, opened 2021-09-10
Switch to GitHub Actions CI

XhmikosR:
`npm test` is basically build right now. I plan to add ESLint and Stylelint later.

Alternatively you could use Cloudflare Pages or Netlify for both PR previews and the main site later.

Preview test run: https://github.com/XhmikosR/svgomg/runs/3567349090?check_suite_focus=true


.github/workflows/ci.yml
```diff
@@ -0,0 +1,62 @@
...
+        uses: actions/download-artifact@v2
+        with:
+          name: docs
+          path: ./build/
+
+      - name: Deploy
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          allow_empty_commit: false
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
XhmikosR, line 60: This should be enough IIRC. If not, we could use a PAT later.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Test
+        run: npm test
+
+      - name: Upload docs
+        uses: actions/upload-artifact@v2
+        if: github.repository == 'jakearchibald/svgomg' && github.ref == 'refs/heads/main'
+        with:
+          name: docs
+          path: ./build/
+          if-no-files-found: error
```
jakearchibald, line 39: What's this doing?

XhmikosR: I basically keep the build files from the test step, upload them as a zip and then download and extract them in the deploy step (instead of rebuilding).

XhmikosR: You can see this in action here for example: https://github.com/twbs/blog/runs/3535610220

.github/workflows/ci.yml
jakearchibald: Where will this be set?

.github/workflows/ci.yml
jakearchibald: Aha, the renaming makes it clearer!

.github/workflows/ci.yml
XhmikosR: `GITHUB_TOKEN` is set by default https://docs.github.com/en/actions/reference/authentication-in-a-workflow#about-the-github_token-secret

If we go with PAT, then in the repo settings -> Secrets.

jakearchibald:
In terms of deployment, can we create a `live` branch for that? That means we can merge PRs that fix features, then when it's time to update the site, we can bump the version, update the changelog etc etc, then advance the live branch.

Also means we can rollback the live branch in case of disaster 😄

XhmikosR:
How about we just use git tags? So, we only deploy on gh-pages when a git tag is made.

jakearchibald:
Doesn't that heavily restrict what tags can be used for? Also, how would rolling back work in that model? Deleting tags?

XhmikosR:
Reverting any offending patches and cutting a new tag :)

Sounds like you are better off using Cloudflare Pages or Netlify and forget about GitHub Pages. The only issue I see with that is the domain change.

jakearchibald:
Yeah, I don't think it's worth the domain change.

I think I prefer using a `live` branch for deployment, since you can say "the commit the live branch points to is the deployed version", whereas with the tag thing it's something like "the commit the most recently created tag points to is…" or "the commit the most recently pushed tag points to is…" or "the commit with the most recent timestamp that has a tag is…" it's pretty ambiguous.

XhmikosR:
I'm not sure I follow your last point...

Can you help me understand what do you want to go to the gh-pages branch and what to the live branch and how we are supposed to distinguish between them?

BTW the deployment action will point to the commit https://github.com/twbs/blog/commits/gh-pages if that helps.

XhmikosR:
I reverted the deployment part so that we move forward, but let me know if I could help with automating deployments.

jakearchibald:
> Can you help me understand what do you want to go to the gh-pages branch and what to the live branch and how we are supposed to distinguish between them?

When the `live` branch changes position, the CI would run `npm run build` against this commit, and the resulting content of the `build` directory would be used in a new commit on the `gh-pages` branch. I'm pretty sure it's the same as what you had before, except it was "when the `main` branch changes position" rather than "when the `live` branch changes position".

You should be able to tell the difference between the two by their different names and content, as we currently do with `main` vs `gh-pages`.

XhmikosR:
OK, so basically you to have two branches for code: one main and one live. Then the deployments to the `gh-pages` branch will happen when there's something merged to the `live` branch. Is that correct?

jakearchibald:
Correct! It means PRs can be merged against master without causing a redeploy of the site. This is useful when combining multiple changes into one release, and separates the task of updating the changelog etc.

XhmikosR:
The last push should probably be what you want :)

You just need to make a live branch too, and perhaps document this somewhere.

jakearchibald (approved):

jakearchibald:
LGTM. Anything need to be done before merging this?

XhmikosR:
Just the 2 things I mentioned above; this will run after the merge. We can fix any issues that might arise :)

jakearchibald:
Ok, let's go!

PR #250, opened 2021-09-10
Fix font after #247.

XhmikosR:
At some point, might be worth switching to the new URL :)

Sorry for the small regression!

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -186,7 +186,7 @@
           requestAnimationFrame(function() {
             [
               'all.css',
-              'https://fonts.googleapis.com/css?family=Roboto:400,700|Inconsolata'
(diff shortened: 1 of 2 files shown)


jakearchibald (approved):

XhmikosR:
True, I thought you wanted to keep using Google Fonts :)

I personally just serve the fonts on my projects. Maybe I'll have a stab at it after the dust settles.

jakearchibald:
> True, I thought you wanted to keep using Google Fonts :)

Haha nah. They don't pay me commission 😄 

> I personally just serve the fonts on my projects.

Yeah, that's what I do on other projects. Back when I started this project, different browsers needed different formats.

PR #253, opened 2021-09-10
Add ESLint scripts and fix issues

XhmikosR:


package.json
```diff
@@ -21,9 +21,11 @@
     "svgo": "^2.5.0"
   },
   "scripts": {
+    "lint": "eslint --report-unused-disable-directives --ignore-path .gitignore .",
+    "fix": "npm run lint -- --fix",
     "build": "gulp clean-build --production",
     "start": "sirv build --port 8080 --dev --no-clear",
     "dev": "gulp dev",
-    "test": "gulp build --production"
+    "test": "npm run lint && gulp build --production"
```
TrySound, line 29: Maybe group lint, fix, test together?

XhmikosR: Done. BTW isn't the `production` flag moot?

TrySound: 🤷‍♂️

XhmikosR: Yup, I confirmed it, the `--production` flag is moot since a long time.

XhmikosR:
@jakearchibald friendly ping since I have a branch that depends on this :)

BTW do you have an ESLint config you want to use? It will help me with the rest of the patches.

jakearchibald (approved):

jakearchibald:
I'm happy with the current rules, although in other projects I've switched away from eslint and just use prettier (with single quotes).

XhmikosR:
prettier is just for formatting, it can't catch real issues...

jakearchibald:
Ah yeah, I'm using Typescript on everything else.

XhmikosR:
So, what do you think we should use here since Typescript is not in use? Do you have a config that matches your preference we could use?

jakearchibald:
Are there configs you'd recommend? I can take a look and see if there's any tweaks needed. Otherwise I can go through the rules and create a custom set.

XhmikosR:
I personally use [xo](https://github.com/xojs/xo) because I don't like reinventing the wheel. It has a `prettier` option we could leverage.

I know Trysound doesn't like it due to the number of dependencies it has, and I agree. But IMHO the pros outweigh the cons.

XhmikosR:
For example, https://github.com/XhmikosR/svgomg/commit/b31010e310d34446ff742c23cb6a45e6ff4fdd28?w=1 (minus the curly and whitespace changes because I didn't use the `prettier` option here)

jakearchibald:
The only stuff I have an opinion on is:

- I like semicolons at the end of lines.
- I prefer `'` to `"` since it's fewer keys.
- I like trailing commas in multiline statements.

I don't mind single-line `if`s for early returns, but I know enough people disagree with that, so I don't mind.

@TrySound any preferences here?

PR #255, opened 2021-09-11
npm scripts: remove the unused `--production` flag

XhmikosR:


jakearchibald:
Hah, I wonder if this ever did anything

XhmikosR:
At some point you were using `env` and different tasks for development and production (which is not a bad idea IMHO as it would speed up development without uglify etc).

PR #258, opened 2021-09-11
Fix logo on Safari

XhmikosR:
I don't own an Apple device so this slipped :/

I just tested on BrowserStack and it seems that without `width`/`height`, the SVG overflows the menu. An alternative way would be to specify the `width`/`height` attributes in the SVG.

@jakearchibald Let me know which solution you prefer.


jakearchibald (approved):
LGTM!

jakearchibald:
This is the solution I'd prefer 😀

Issue #252, opened 2021-09-10
Update Prism/revisit rollup fixes


XhmikosR:
Not sure if this is still needed with the latest prismjs. There's also a build warning:

```
Sourcemap is likely to be incorrect: a plugin (rollup-fixes) was used to transform files, but didn't generate a sourcemap for the transformation. Consult the plugin documentation for help
```

BTW I was wondering, shouldn't we load the Prism languages we only need?

XhmikosR:
The included languages should be OK as it is.

Another thing worth revisiting is the Prism hack. There must be another way to do things without overriding the `message` listener.

jakearchibald:
Fixed in https://github.com/jakearchibald/svgomg/pull/218

PR #259, opened 2021-09-12
Remove prism fix in rollup config

TrySound:
Ref https://github.com/jakearchibald/svgomg/issues/252

Looks like this fix is no longer necessary. Probably was a side effect
of something else.

cc @XhmikosR

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -59,15 +59,6 @@ async function html() {
 
 const rollupCaches = new Map();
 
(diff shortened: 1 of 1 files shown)


XhmikosR:
Yeah, I also came to the same conclusion from my testing; it seems it works fine.

jakearchibald (approved):
Nice! An earlier version or Prism didn't do modules properly.

PR #260, opened 2021-09-12
Replace events with nanoevents

TrySound:
There are a few concerns about "events"

- it's bulitin node module and confuses bundlers
- not well maintained
- got size increase in recent version (https://bundlephobia.com/package/events@3.3.0)
- inheritance makes it kinda implicit

NanoEvents package is very small (72 bytes minified and gzipped) and
written in es modules which allows to avoid any commonjs related size increases.


jakearchibald:
I wonder if extending `EventTarget` is supported enough now. If not, I'm pretty sure `new EventTarget` is supported enough.

TrySound:
EventTarget api is a bit messy comparing to both events and nanoevents. Need to create some event object, somehow pass the data.

jakearchibald:
Fwiw, this would be the equivalent, but using event targets:

```js
class EventEmitter extends EventTarget {
  emit(type, detail) {
    this.dispatchEvent(new CustomEvent(type, { detail }));
  }
  on(type, callback) {
    const wrappedCallback = (event) => callback(event.detail);
    this.addEventListener(type, wrappedCallback);
    return () => this.removeEventListener(type, wrappedCallback);
  }
}
```

…but nanoevents is smaller, so there's no benefit using actual event targets.

jakearchibald (approved):

PR #261, opened 2021-09-13
Global clipboard events - copy and paste

### LINKED ISSUE #149: Enable pasting SVG without clicking the "Paste markup" button
In Figma, if you have SVG code in your clipboard and just hit Ctrl+V (or Cmd+V on macOS) in a document, it creates a new frame with your SVG in that document. It's a really nice way to work with SVG quickly.

It would be great to have the same in SVGOMG. It probably needs no UI. Only need to listen for the right keyboard shortcut on the page. Though I guess there might be security restrictions that Figma works around with an invisible input, I’m not sure.

gismya:
Added global copy and paste events. Ctrl+c copies the current SVG markup (Like the "Copy to text-button") and ctrl+v now globally works like the paste markup-box.

Fixes #149


XhmikosR:
There should at least be some kind of visual indication IMHO. Maybe use a Toast?

src/js/page/main-controller.js
```diff
@@ -128,6 +130,39 @@ export default class MainController {
     }
   }
 
+  _onGlobalPaste(event) {
+    let val = event.clipboardData.getData('text').trim();
```
jakearchibald, line 134:
```suggestion
    const val = event.clipboardData.getData('text').trim();    
```

src/js/page/main-controller.js
```diff
@@ -128,6 +130,39 @@ export default class MainController {
...
+    if (!val.includes("</svg>")) {
+      this._toastsUi.show("Pasted value not an SVG", {
+        duration: 2000
+      });
+    }
+    else {
+      this._mainMenuUi.setPasteInput(val);
+      this._toastsUi.show("Paste completed", {
+        duration: 2000
+      });
```
jakearchibald, line 144:
```suggestion
```

The toast is great for copy, but I don't think it's needed for paste. The UI already makes it clear that something happened.

src/js/page/ui/copy-button.js
```diff
@@ -23,6 +23,12 @@ export default class CopyButton extends FloatingActionButton {
 
   _onClick(event) {
     super._onClick(event);
+    this.getCopyText();
+  }
+
+  getCopyText() {
```
jakearchibald, line 29: Could you rename this to indicate it performs an action? Right now it reads like it "gets the text to be copied". Maybe `copyText`?

jakearchibald (changes requested):
Almost there! Nice feature

src/js/page/ui/copy-button.js
```diff
@@ -35,9 +41,12 @@ export default class CopyButton extends FloatingActionButton {
...
     getSelection().removeAllRanges();
     document.body.removeChild(this._pre);
+    
+    return true;
   }
-
   setCopyText(text) {
     this._text = text;
   }
+
```
XhmikosR, line 50: Nitpicking these 2 newlines are not needed :)

src/js/page/ui/main-menu.js
```diff
@@ -57,6 +57,11 @@ export default class MainMenu {
     this._loadFileInput.click();
   }
 
+  setPasteInput(value) {
+    this._pasteInput.value = value;
+    this._pasteInput.dispatchEvent(new Event('input'));
```
XhmikosR, line 62: Not sure about using `Event`; `nanoevents` are used everywhere else.

jakearchibald: It's a little bit of a hack, but if an event is going to be used here, it's the right kind of event, since this is an `HTMLElement`.

jakearchibald:
@TrySound @XhmikosR I'm happy to merge this as-is and create a seperate ticket for a bit of refactoring. Or do you think that should be done before merging?

Feels a bit weird having the controller call into the copy button when keys are pressed. Instead, the copy functionality should be moved into the controller or a utility, and the controller calls it when the button is pressed, or when the keys are pressed.

Similarly with paste, right now it's calling into the menu to set the paste input and manufacturing an input event. Seems like this should be moved a level up.

XhmikosR:
The only thing I think it makes sense to have before this lands is a consistent usage of `copySupported`.

The rest I leave it to you guys :)

src/js/page/ui/copy-button.js
gismya: This is why people invented automatic linting ;)

jakearchibald:
I don't think we need `copySupported` anymore, since all modern browsers support it. @gismya do you mind removing the usage of that as part of this PR? No worries if not, I'm happy to do it.

TrySound (approved):

gismya:
@jakearchibald - I agree that going into the UI components to access the functions felt extremely hacky - but didn't want to come in and refactor too much "uninvited". I feel an utility would be the most suitable place. I can look into making those changes tonight probably.

.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
```
TrySound, line 1: I would not commit vscode config here

gismya:
Haha, I'm too slow with reading all the updated. I'll remove copySupported

.vscode/launch.json
gismya: Fudge - mistake from my part. Managed to remember to remove it in every commit until now. My bad!

I'll add the .vscode-folder to gitignore :)

TrySound: Maybe ignore it globally?

gismya: Did not know I could have a global gitignore. Now that's set up :)

gismya:
Didn't meditate too long on the refactor, so I apologize in advance for any breaches of code conventions, but I made another PR for my proposed refactoring: https://github.com/gismya/svgomg/pull/1

jakearchibald (approved):
This is great, thanks so much for adding this!

PR #265, opened 2021-09-14
Avoid UI obscuring code

jakearchibald:
Before:

jakearchibald github io_svgomg_ (1)

After:

localhost_8080_ (1)

### DIFF
--- a/src/css/components/_code-output.scss
+++ b/src/css/components/_code-output.scss
@@ -25,6 +25,11 @@
     font-size: 0.9rem;
     display: block;
     padding: 14px;
+
+    @media (min-width: 640px) {
+      // Allow the code area to scroll past the floating action buttons.
+      margin-bottom: 200px;
+    }
   }
 
   @media (min-width: 900px) {
--- a/src/css/_main-layout.scss
+++ b/src/css/_main-layout.scss
@@ -11,7 +11,7 @@
 
   @media (min-width: 640px) {
     flex: none;
-    min-width: 21.1em;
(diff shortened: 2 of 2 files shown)


TrySound (approved):

src/css/components/_code-output.scss
```diff
@@ -25,6 +25,11 @@
     font-size: 0.9rem;
     display: block;
     padding: 14px;
+
+    @media (min-width: 640px) {
+      // Allow the code area to scroll past the floating action buttons.
+      margin-bottom: 200px;
```
XhmikosR, line 31: @jakearchibald I was thinking, maybe 200px is too much?

jakearchibald: Hm, it can't be much less before the code goes under the background-change icon.

jakearchibald: It looks like there's a lot of space in the screenshot above, but the text isn't all the way across on the final line.

XhmikosR: Yeah, that's what I meant. Not sure why this happens yet.

XhmikosR:
@jakearchibald Another thing I notice after this PR is that when cache is empty, after clicking on Demo, the transition is a bit sluggish. It's like something is firing up at the same time; maybe the loader and something else?

jakearchibald:
I'm pretty sure it's not this PR causing it. The changed style only resolves when you go to the code view, which isn't the default view.

PR #268, opened 2021-09-14
Close main menu when the Escape key is pressed

XhmikosR:
Not 100% familiar with the codebase but I think this is the place this should be. If not let me know :)


jakearchibald (approved):
LGTM!

PR #269, opened 2021-09-14
Use SVGO multipass

jakearchibald:
I removed the 'custom' multipass & now just use SVGO's. I also made aborting now terminate the worker, so it's an instant abort rather than 'next iteration'.


TrySound (approved):
LGTM. Hope eventually multipass will not be necessary.

Issue #254, opened 2021-09-11
Remove old browsers workarounds


XhmikosR:
The codebase has some IE workarounds and maybe old Safari ones. Would be good to clean them up.

PR #271, opened 2021-09-14
Remove IE leftovers

XhmikosR:
Basically reverts 027fd5abd29c080761f3259e63f8933b113275d1 and 6a9a33ec5248416eb918112a7843456b99e80ac9

I haven't tested this on Safari; Edge and Firefox work fine AFAICT.

Refs #254

### DIFF
--- a/src/css/components/_main-menu.scss
+++ b/src/css/components/_main-menu.scss
@@ -82,8 +82,7 @@
   & .menu-item {
     padding: 0 16px;
     width: 100%;
-    // I want this to be min-height, but IE breaks layout if I do
-    height: 48px;
+    min-height: 48px;
     box-sizing: border-box;
     display: flex;
     align-items: center;
--- a/src/js/page/ui/material-slider.js
+++ b/src/js/page/ui/material-slider.js
@@ -42,7 +42,6 @@ export default class MaterialSlider {
     this.range.classList.add('active');
 
     const upListener = () => {
-      // IE requires me to do this. Pah.
       requestAnimationFrame(() => {
         this.range.blur();
       });
--- a/src/css/components/_material-slider.scss
+++ b/src/css/components/_material-slider.scss
@@ -58,23 +58,11 @@
     position: absolute;
     top: 0;
     left: 0;
-    margin: 0;
-    width: 100%;
-
-    // increase the touch target
-    height: 60px;
-
-    @supports (width: calc(100% + 10px)) {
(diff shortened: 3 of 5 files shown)


src/js/page/ui/material-slider.js
```diff
@@ -42,7 +42,6 @@ export default class MaterialSlider {
     this.range.classList.add('active');
 
     const upListener = () => {
-      // IE requires me to do this. Pah.
```
XhmikosR, line 45: I removed this comment because it appears to be incorrect; Without the next call, I saw breakage on at least Firefox.

src/css/components/_main-menu.scss
```diff
@@ -82,8 +82,7 @@
   & .menu-item {
     padding: 0 16px;
     width: 100%;
-    // I want this to be min-height, but IE breaks layout if I do
-    height: 48px;
+    min-height: 48px;
```
XhmikosR, line 85: @jakearchibald I went with what you wanted in the comment. That being said, not 100% sure this doesn't cause a small reflow. Assuming the menu items can never be taller than 48px, I'd personally stick with `height` :)

jakearchibald: `min-height` seems better to me, but it doesn't massively matter with the current UI.

Before:

jakearchibald github io_svgomg_ (2)

After:

deploy-preview-271--affectionate-lovelace-6ea273 netlify app_

XhmikosR: Yeah, I know that, but with the current UI it's not needed is my point. Anyway, your call; if you don't see any reflows or anything bad, we can go with `min-height`, otherwise I can switch back to `height`

src/js/page/ui/material-slider.js
jakearchibald: Hah IE was right! Yeah other browsers were getting the event loop wrong here.

jakearchibald (approved):
LGTM! Thanks

src/css/components/_main-menu.scss
jakearchibald: `min-height` shouldn't cause a visual reflow here, since it'll know the height of each item in a single layout.

XhmikosR: Though, I do personally see a reflow when Ctrl+F5. Maybe reflow isn't the right term, but the `.menu-item` is resized. Using the latest Edge/Chrome.

PR #273, opened 2021-09-14
view-toggler.js: remove workaround

### LINKED ISSUE #254: Remove old browsers workarounds

XhmikosR:
I tested this and I don't see any issues. Please confirm.

Closes #254 since I believe this is the last old browser workaround.


jakearchibald (approved):
Any change that removes `array.reduce` gets a 👍 from me

PR #274, opened 2021-09-14
Improve gulpfile.js

XhmikosR:
* fix duplicate sass call
* use `path.join` when dealing with file paths
* only copy the needed files in the build dir
* fix watch for json files
* clean on dev too

Backporting some of my branch changes.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -2,12 +2,6 @@ const fs = require('fs/promises');
 const path = require('path');
 const sirv = require('sirv-cli');
 const svgoPkg = require('svgo/package.json');
-
-const readJSON = async (path) => {
-  const content = await fs.readFile(path, 'utf-8');
-  return JSON.parse(content);
-};
-
 const sass = require('sass');
 const gulp = require('gulp');
 const gulpSourcemaps = require('gulp-sourcemaps');
@@ -20,22 +14,25 @@ const rollupCommon = require('@rollup/plugin-commonjs');
 const rollupReplace = require('@rollup/plugin-replace');
 const { terser: rollupTerser } = require('rollup-plugin-terser');
 
+const readJSON = async path => {
+  const content = await fs.readFile(path, 'utf-8');
+  return JSON.parse(content);
+};
+
 function css() {
   const boundSass = gulpSass(sass);
-  return gulp.src('src/css/*.scss', { sourcemaps: true })
-    .pipe(boundSass.sync().on('error', boundSass.logError))
+  return gulp.src('src/css/*.scss')
     .pipe(gulpSourcemaps.init())
-    .pipe(boundSass({ outputStyle: 'compressed' }))
+    .pipe(boundSass.sync({ outputStyle: 'compressed' }).on('error', boundSass.logError))
     .pipe(gulpSourcemaps.write('./'))
     .pipe(gulp.dest('build/'));
 }
 
-
 async function html() {
   const [config, changelog, headCSS] = await Promise.all([
-    readJSON(`${__dirname}/src/config.json`),
-    readJSON(`${__dirname}/src/changelog.json`),
-    fs.readFile(`${__dirname}/build/head.css`)
+    readJSON(path.join(__dirname, 'src', 'config.json')),
+    readJSON(path.join(__dirname, 'src', 'changelog.json')),
+    fs.readFile(path.join(__dirname, 'build', 'head.css'), 'utf-8')
   ]);
 
   return gulp.src('src/*.html')
@@ -67,9 +64,8 @@ async function html() {
 const rollupCaches = new Map();
 
 async function js(entry, outputPath) {
-  const parsedPath = path.parse(entry);
-  const name = /[^/]+$/.exec(parsedPath.dir)[0];
-  const changelog = await readJSON(`${__dirname}/src/changelog.json`);
+  const name = path.basename(path.dirname(entry));
+  const changelog = await readJSON(path.join(__dirname, 'src', 'changelog.json'));
   const bundle = await rollup.rollup({
     cache: rollupCaches.get(entry),
     input: `src/${entry}`,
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -20,30 +14,33 @@ const rollupCommon = require('@rollup/plugin-commonjs');
...
 const { terser: rollupTerser } = require('rollup-plugin-terser');
 
+const readJSON = async path => {
+  const content = await fs.readFile(path, 'utf-8');
+  return JSON.parse(content);
+};
+
 function css() {
   const boundSass = gulpSass(sass);
-  return gulp.src('src/css/*.scss', { sourcemaps: true })
```
XhmikosR, line 25: Removed `{ sourcemaps: true }` since we are loading only scss files which don't have sourcemaps

gulpfile.js
```diff
@@ -94,45 +91,28 @@ const allJs = gulp.parallel(
...
-    'src/.**/*',
-    // Exclude the following files
-    // (other tasks will handle the copying of these files)
-    '!src/*.html',
-    '!src/{css,css/**}',
-    '!src/{js,js/**}'
+    'src/{.well-known,imgs,test-svgs}/**',
+    // Exclude the test-svgs files except for `car-lite.svg`
+    // which is used in the demo
+    '!src/test-svgs/!(car-lite.svg)',
```
XhmikosR, line 97: I had to use this weird-looking expression to get to keep only car-lite.svg

gulpfile.js
```diff
@@ -20,30 +14,33 @@ const rollupCommon = require('@rollup/plugin-commonjs');
...
     .pipe(gulp.dest('build/'));
 }
 
-
 async function html() {
   const [config, changelog, headCSS] = await Promise.all([
-    readJSON(`${__dirname}/src/config.json`),
-    readJSON(`${__dirname}/src/changelog.json`),
-    fs.readFile(`${__dirname}/build/head.css`)
+    readJSON(path.join(__dirname, 'src/config.json')),
```
jakearchibald, line 33: Should these use `path.join` for all parts of the path? As in `path.join(__dirname, 'src', 'config.json')`?

gulpfile.js
jakearchibald: We can just remove the other test SVGs to avoid this

gulpfile.js
XhmikosR: It's the same and personally I prefer the shorter way :)

gulpfile.js
XhmikosR: Your call, I find some of the other test files good to have, but now they won't be deployed anyway. Maybe just keep them for historical reasons?

XhmikosR: Another solution would be to output the demo SVG to the build folder; that would also get rid of this workaround.

BTW the demo SVG could also probably be in the SW; it's not that big and would probably simplify some of the JS code.

gulpfile.js
jakearchibald: I'm missing something here… I thought the purpose of `path.join` was to cater for platform differences.

If platform differences don't matter, then `__dirname + '/src/config.json'` is fine.

If platform differences matter, then you'd need `path.join(__dirname, 'src', 'config.json')` surely?

I can't figure out where `path.join(__dirname, 'src/config.json')` fits in this.

gulpfile.js
```diff
@@ -67,12 +64,11 @@ async function html() {
...
 async function js(entry, outputPath) {
-  const parsedPath = path.parse(entry);
-  const name = /[^/]+$/.exec(parsedPath.dir)[0];
-  const changelog = await readJSON(`${__dirname}/src/changelog.json`);
+  const name = path.basename(path.dirname(entry));
+  const changelog = await readJSON(path.join(__dirname, 'src/changelog.json'));
   const bundle = await rollup.rollup({
     cache: rollupCaches.get(entry),
-    input: `src/${entry}`,
+    input: path.normalize(`src/${entry}`),
```
jakearchibald, line 71: Doesn't rollup handle this already?

gulpfile.js
XhmikosR: Because `path.join(__dirname, 'src/config.json')` already fixes any platform differences regardless of the number of parts :)

gulpfile.js
XhmikosR: Probably, yeah. I can revert this and the `file` change.

gulpfile.js
jakearchibald:
```js
path.join('c:\\foo\\bar', 'src/config.json');
```

…is `c:\foo\bar/src/config.json`, creating a mix of path separators.

XhmikosR:
```js
const path = require('path');

let p = __dirname + '/src/config.json';
console.log(`p: ${p}`); // prints: p: C:\Users\xmr\Desktop/src/config.json

p = path.join(__dirname, 'src', 'config.json');
console.log(`p: ${p}`); // prints: p: C:\Users\xmr\Desktop\src\config.json

p = path.join(__dirname, 'src/config.json');
console.log(`p: ${p}`); // prints: p: C:\Users\xmr\Desktop\src\config.json
```

XhmikosR:
```js
console.log(path.join('c:\\foo\\bar', 'src/config.json')); // prints: c:\foo\bar\src\config.json
```

jakearchibald: Ah I see, it's because it normalises the path afterwards, and on Windows that means swapping `/` for `\`. On Posix:

```js
console.log(path.join('c:\\foo\\bar', 'src/config.json')); // prints: c:\foo\bar/src/config.json
```

I realise it works because the above will never happen, as the first half will always be Posix on Posix systems, but it feels like a bit of a hack. I'd prefer to use `path.join` for all parts of the path, except the parts within `__dirname` of course as we already know they're in the platform's format.

XhmikosR: Well, that's because on Posix systems `\` is allowed to be in paths, i.e. is not a separator.

I think we are fine here with the patch as is, but if you insist I can change it, although I don't think it's needed.

jakearchibald: My problem with it is in works in this one situation, but it doesn't work as a general rule, and requires particular knowledge to know why it's ok. I'd rather go with `path.join(__dirname, 'src', 'config.json')` so it isn't confusing in future.

@TrySound I think we need a tie-break opinion here.

XhmikosR: I use your solution, either way it's better

gulpfile.js
jakearchibald: Happy to land this as-is. Can always split this into `demo.svg` and `test-svgs` later (where the latter isn't part of the build).

jakearchibald (approved):
Thanks for putting up with the back & forth on this!

PR #278, opened 2021-09-15
Refactored loading and clipboard functions

gismya:
Not quite sure about the naming, but here's an idea on how to structure the clipboard and loading functions.

### DIFF
--- a/src/js/page/utils.js
+++ b/src/js/page/utils.js
@@ -1,3 +1,5 @@
+import { createNanoEvents } from 'nanoevents';
+
 export const domReady = new Promise(resolve => {
   function checkState() {
     if (document.readyState != 'loading') resolve();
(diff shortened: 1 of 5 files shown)


src/js/page/utils.js
```diff
@@ -89,3 +91,42 @@ export function trackFocusMethod() {
...
+
+    window.getSelection().addRange(range);
+
+    document.execCommand('copy');
+    getSelection().removeAllRanges();
+    document.body.removeChild(this._pre);
+    
+    return true;
+  }
+};
```
jakearchibald, line 118: Could this just be `copyText(text)`? Instead of calling `setCopyText()`, components would just store the text for the  SVG (they probably already do).

src/js/page/utils.js
```diff
@@ -89,3 +91,42 @@ export function trackFocusMethod() {
...
+      this.loadSvg(val, "image.svg");
+    }
+  },
+  loadSvg(data, filename) {
+    this.emitter.emit("svgDataLoad", {
+      data,
+      filename,
+    });
+  },
+};
```
jakearchibald, line 132: Similar to above, I think there's some indirection here. Both of these could be methods in the main controller, removing the need for the event.

PR #283, opened 2021-09-17
package.json: add missing properties

XhmikosR:
This is mostly nitpicking since the package isn't published but `npm init -y` adds them. Feel free to pass if you think it's not worth it.


jakearchibald (approved):
Why not 😄

PR #284, opened 2021-09-17
utils/storage.js: return early

XhmikosR:


src/js/utils/storage.js
```diff
@@ -2,25 +2,24 @@ export let idbKeyval = (() => {
   let db;
 
   function getDB() {
-    if (!db) {
-      db = new Promise((resolve, reject) => {
-        const openreq = indexedDB.open('svgo-keyval', 1);
+    if (db) return db;
```
TrySound, line 5: So db will never be cached?

TrySound: I'd say the code is fine as is. Extra nesting is not always bad.

jakearchibald: I generally prefer early returns, but yeah, when it comes to caching I prefer:

```js
function whatever() {
  if (notCached) addToCache();
  return getFromCache();
}
```

To:

```js
function whatever() {
  if (cached) return getFromCache();
  addToCache();
  return getFromCache();
}
```

The latter feels more prone to error, like changing how one of the returns works, or forgetting the `addToCache` bit (which is what's happening in this PR).

jakearchibald:
Going to close this one as I don't think it needs changing. We could switch to [idb-keyal](https://github.com/jakearchibald/idb-keyval) at some point, but doesn't really seem necessary.

XhmikosR:
To clarify, my intention was to keep setting the `db` variable. I accidentally removed the assignment.

PR #285, opened 2021-09-17
material-slider.js: switch to plain strings

XhmikosR:
This is minified better without the extra whitespace


jakearchibald (approved):
This is where things like JSX comes in useful I guess. I stopped myself from starting to rewrite the site in Preact earlier 😄

PR #286, opened 2021-09-17
index.html: remove redundant `requestAnimationFrame` call

XhmikosR:
@jakearchibald I *think* this is redundant, but maybe I'm missing something :)

BTW would be nice to switch to `link` and `script` tags instead of the inline JS, assuming the outcome is the same.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -190,27 +190,25 @@
         });
 
         requestAnimationFrame(function() {
-          requestAnimationFrame(function() {
-            [
-              'all.css',
-              'https://fonts.googleapis.com/css?family=Roboto:400,700%7CInconsolata'
-            ].forEach(function(url) {
-              var link = document.createElement('link');
-              link.rel = 'stylesheet';
-              link.crossorigin = true;
-              link.href = url;
+          [
+            'all.css',
+            'https://fonts.googleapis.com/css?family=Roboto:400,700%7CInconsolata'
+          ].forEach(function(url) {
+            var link = document.createElement('link');
+            link.rel = 'stylesheet';
+            link.crossorigin = true;
+            link.href = url;
 
-              document.head.appendChild(link);
-            });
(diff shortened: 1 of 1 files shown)


jakearchibald:
> @jakearchibald I _think_ this is redundant, but maybe I'm missing something :)

This was originally forcing the browser to perform an early initial render to avoid parsing the rest of the script. Browser behaviour may have changed here, but it'll need testing.

> BTW would be nice to switch to `link` and `script` tags instead of the inline JS, assuming the outcome is the same.

It'll be slower, which isn't great. What's the benefit of switching?

XhmikosR:
> This was originally forcing the browser to perform an early initial render to avoid parsing the rest of the script. Browser behaviour may have changed here, but it'll need testing.

Let me know what you find out, I knew there was a reason but didn't know what it was 😛

> It'll be slower, which isn't great. What's the benefit of switching?

If we use scirpt async or defer along with preloading, wouldn't we have the same result?

jakearchibald:
> If we use scirpt async or defer along with preloading, wouldn't we have the same result?

Nah, making a request is still slower than not making a request. It's why we inline first-interaction script and CSS on squoosh.app too.

jakearchibald:
Oh wait, ignore me, we're talking about different things.

It'd be best to inline the JS needed for first interaction, but I agree that we should try to use simple tags for preloading here.

XhmikosR:
> It'd be best to inline the JS needed for first interaction, but I agree that we should try to use simple tags for preloading here.

Yeah, that was my point :)

I'll take a stab at it later or open a new issue so that we don't forget about it.

jakearchibald:
I'm having a play in https://github.com/jakearchibald/svgomg/pull/289

XhmikosR:
Perfect, thanks! BTW I just find it too hard to judge on Netlify because hitting their cache is like rolling a dice :/

PR #289, opened 2021-09-17
Improving preloading

jakearchibald:
Draft while I check this doesn't regress performance.

Before: https://www.webpagetest.org/result/210917_AiDcE0_ea85f387b84f1f2e0fc2f5506826cf18/
After: …

Closes #286 by superseding it


jakearchibald:
Ok, I remembered why I was doing everything in script, it was to avoid keeping the browser spinner going. I think I have a better solution now, but waiting for tests to complete.

jakearchibald:
This doesn't delay the browser spinner, but still gets things loading early. Seems like a decent system?

src/index.html
```diff
@@ -182,37 +199,13 @@ <h1 class="settings-sub-heading">Features</h1>
...
-
-        requestAnimationFrame(function() {
-          requestAnimationFrame(function() {
-            [
-              'all.css',
-              'https://fonts.googleapis.com/css?family=Roboto:400,700%7CInconsolata'
-            ].forEach(function(url) {
-              var link = document.createElement('link');
-              link.rel = 'stylesheet';
-              link.crossorigin = true;
```
jakearchibald, line 200: Hah, this was doing nothing, since it's `crossOrigin` not `crossorigin`.

XhmikosR: See, it's good fiddling with things that haven't been touched for a long time, after all! 😛

src/index.html
```diff
@@ -182,37 +199,13 @@ <h1 class="settings-sub-heading">Features</h1>
         </div>
         <p>Sorry, wasn't ready…</p>
       </div>
+      <link rel="preload" as="script" href="js/page.js" onload="appendScript(href)">
```
XhmikosR, line 202: Can you elaborate a bit why you move this `link` here and not in `head`?

I'm sure there's a reason.

XhmikosR:
It works pretty good AFAICT, though you know this sh*t a lot better than I do :)

src/index.html
jakearchibald: To ensure the script doesn't load before the DOM is present

CyberAP: Is it better than regular `link preload` in head and then `script defer` before body closing tag? I assumed it would unnecessry trigger DOM recalculations, while the old approach theoretically could avoid that. Was it wrong to assume?
Is this done to achieve a higher priority at a cost of a later request?

jakearchibald: `script defer` delays the browser loading spinner, which is what I'm trying to avoid.

src/index.html
```diff
@@ -5,11 +5,28 @@
...
+        link.rel = 'stylesheet';
+        // So it can be cached without no-cors in the service worker:
+        link.crossOrigin = true;
+        link.href = url;
+        document.head.appendChild(link);
+      }
+    </script>
+    <link rel="preload" crossorigin as="style" href="all.css" onload="appendStyle(href)">
+    <link rel="preload" crossorigin as="style" href="https://fonts.googleapis.com/css?family=Roboto:400,700%7CInconsolata" onload="appendStyle(href)">
+    <link rel="preconnect" crossorigin href="https://fonts.googleapis.com">
```
rik, line 28: Why preconnect to a domain used one line above? That should have no effect, right?

Are you trying to preconnect to the domain Google Fonts uses for the actual font files? aka https://fonts.gstatic.com

jakearchibald: Yes, I am. Good catch!

kurtextrem:
Question: Isn't the `appendScript` kind of cheating in terms of UX (was the same with the double rAF)? 
So to say the app is loading, while the loading spinner is no longer spinning. Or am I missing a point? (for GA I understand it, since it's not user facing, but page.js is needed for functionality - or not?)

jakearchibald:
@kurtextrem depends what you mean by functionality. There's code to handle a too-early-click, but that rarely ever happens.

kurtextrem:
@jakearchibald Yeah, I see, but then again - is showing the "Sorry, wasn't ready" nicer UX than seeing the loading spinner for a few seconds longer?

jakearchibald:
@kurtextrem that's fair. I'll tweak it so those continue to trigger the browser spinner.

XhmikosR (approved):

PR #292, opened 2021-09-20
Add Stylelint and Sass hygiene

XhmikosR:
@jakearchibald not sure if you like the rules I used like the remove leading zero. Or being explicit with some properties (personally I prefer it, but I can revert).

I tried to keep the patches separate so that we can make changes if needed.

PS. Better check the non-whitespace diff: https://github.com/jakearchibald/svgomg/pull/292/files?w=1


src/css/_global.scss
```diff
@@ -1,32 +1,32 @@
...
 pre,
 code,
 kbd,
 samp,
-tt{
-  font-family:monospace,monospace;
-  font-size:1em;
+tt {
+  // stylelint-disable-next-line font-family-no-duplicate-names
+  font-family: monospace, monospace;
```
XhmikosR, line 24: This was needed is some older browsers, not sure if it's still needed...

src/css/_global.scss
```diff
@@ -1,32 +1,32 @@
...
   height: 100%;
   margin: 0;
   padding: 0;
   overflow: hidden;
 }
 
 html {
-  font-family: roboto, sans-serif;
-  -webkit-tap-highlight-color: rgba(0,0,0,0);
+  font-family: Roboto, sans-serif;
```
XhmikosR, line 10: I changed the font case because I couldn't find it at first with search. Otherwise this has no effect.

src/css/_main-layout.scss
```diff
@@ -6,19 +6,34 @@
 }
 
 .settings {
-  flex: 0.9 1 0;
+  flex: .9 1 0;
   overflow: hidden;
+  transform: translateY(110%);
```
jakearchibald, line 11: I'm not sure about this. Before this PR, the bits of CSS relating to the "before file open" state are all in one place, since they're relating to the same conceptual state. But the linter has scattered them throughout the file.

XhmikosR: It's this commit: https://github.com/jakearchibald/svgomg/pull/292/commits/1808d49f98ff3f8cd120b0c3d297975128a316ea

I could revert it, it's just 3 duplicate selectors.

XhmikosR: Or maybe you could see if you could refactor the code the way you want it without the extra selectors?

Either way I can revert :)

src/css/components/_changelog.scss
```diff
@@ -1,31 +1,31 @@
 .changelog {
-  font-size: 0.9rem;
-  color: #898989;
+  font-size: .9rem;
+  color: #767676;
```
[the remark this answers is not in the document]
XhmikosR: Nope, check out the different commits, I kept them separate for this reason 😛

It's so that the contrast ratio it's 4.5:1 for WCAG 2.

src/css/components/_settings.scss
```diff
@@ -60,9 +57,9 @@
 .settings-sub-heading {
   margin: 0;
   padding: 12px 16px 16px;
-  font-weight: normal;
-  font-size: 0.9rem;
-  color: #898989;
+  font-weight: 400;
+  font-size: .9rem;
+  color: #767676;
```
jakearchibald, line 62: Another color change here

src/css/_main-layout.scss
jakearchibald: Yeah, I think revert that one. The duplicate selectors are deliberate to make things easier to maintain. Happy for them be minified away, but don't think it should happen in the file humans are editing.

XhmikosR: Cool, I'll revert this patch. Currently we don't have CSS minification, just whitespace cleanup and generally I'm not sure it's worth it, but I have a patch which adds clean-css with sane settings. We can discuss this later.

.stylelintrc
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": [
+    "stylelint-config-twbs-bootstrap/scss"
+  ],
+  "rules": {
+    "color-named": null,
```
XhmikosR, line 6: @jakearchibald Do you prefer having this off? I fixed white and black since they are easy ones, but there are 3 more instances of named colors so I thought I'd leave this off.

src/css/_main-layout.scss
```diff
@@ -93,83 +93,80 @@
...
 
-    &.transition {
-      transition: transform 0.2s ease-out;
-    }
+  &.transition {
+    transition: transform .2s ease-out;
+  }
 
-    &.active {
-      // translateZ required to make Safari layer in the right order
```
XhmikosR, line 112: I removed these comments because it seems this is needed for at least Firefox too

src/css/components/_material-switch.scss
```diff
@@ -5,69 +5,67 @@
...
     top: -3px;
     left: 0;
     width: 20px;
     height: 20px;
     border-radius: 20px;
-    background: #fafafa;
-    box-shadow: 0 1px 3px rgba(0,0,0,0.5), 
-                inset 0 0 0 2px transparent;
+    background-color: #fafafa;
+    box-shadow: 0 1px 3px rgba(0, 0, 0, .5), inset 0 0 0 2px transparent;
```
XhmikosR, line 41: BTW I wouldn't mind keeping these but I couldn't figure out the proper stylelint combination. If someone else figures it out, let me know.

src/css/components/_settings.scss
```diff
@@ -1,43 +1,40 @@
...
-  overflow-y: auto;
+  overflow: hidden auto;
   -webkit-overflow-scrolling: touch;
   flex: 1;
   transform: translateZ(0);
 }
 
 .setting-item-toggle {
   position: relative;
-  display: block;
```
XhmikosR, line 18: This was duplicate

src/css/components/_spinner.scss
```diff
@@ -1,16 +1,18 @@
...
 @keyframes fill-unfill-rotate {
-  12.5% { transform: rotate(135deg);  } /* 0.5 * ARCSIZE */
-  25%   { transform: rotate(270deg);  } /* 1   * ARCSIZE */
-  37.5% { transform: rotate(405deg);  } /* 1.5 * ARCSIZE */
-  50%   { transform: rotate(540deg);  } /* 2   * ARCSIZE */
-  62.5% { transform: rotate(675deg);  } /* 2.5 * ARCSIZE */
-  75%   { transform: rotate(810deg);  } /* 3   * ARCSIZE */
-  87.5% { transform: rotate(945deg);  } /* 3.5 * ARCSIZE */
-  to    { transform: rotate(1080deg); } /* 4   * ARCSIZE */
+  12.5% { transform: rotate(135deg); } // 0.5 * ARCSIZE
```
XhmikosR, line 8: Changed the comments to `//` so that they are not output in the CSS. With the `compressed` style option this doesn't change anything, though.

.stylelintrc
jakearchibald: Agreed. It seems like a weird rule to me.

src/css/_components.scss
```diff
@@ -1,16 +1,16 @@
...
-@import 'components/material-slider';
-@import 'components/drop-overlay';
-@import 'components/changelog';
-@import 'components/svg-output';
-@import 'components/material-tabs';
-@import 'components/results';
-@import 'components/output';
-@import 'components/code-output';
-@import 'components/ripple';
+@import "components/settings";
```
jakearchibald, line 1: How do you feel about sticking with single quotes like the JS? One key seems better than two 😄

XhmikosR: I personally prefer double quotes in CSS/Sass, but it's your call, I can switch them.

BTW in the past I've hit a pretty nasty IE bug, where using single quotes in a data URI broke IE from rendering.

XhmikosR: https://github.com/mdo/github-buttons/commit/db9e024f389df3e92b6c067becfec32953495048

XhmikosR:
If we add clean-css later like I have on my xmr-dev branch, I could revert the leading zero removal and we either enforce it or not.

jakearchibald:
I'm not excited about the leading zero removal, but I'm not losing any sleep over it. Do whichever is easiest.

I'd still prefer single quotes, as they're easier to type and consistent with the CSS. I totally get that single quotes may have caused IE issues in the past, but that isn't a problem for this project.

Happy for this to merge with those changes.

XhmikosR:
I've already switched to single quotes :)

I can revert the leading zero changes too if you want and then I will rebase the branch.

jakearchibald (approved):
LGTM

PR #293, opened 2021-09-20
Create .editorconfig

XhmikosR:


TrySound:
Does it add any value when prettier is used?

XhmikosR:
Not every file is using prettier and this is the standard to control various aspects of files through various editors.

jakearchibald (approved):

PR #298, opened 2021-09-20
Move SVG icons outside of the HTML

XhmikosR:
I think it makes sense to do this and I personally find this cleaner :)

PS. the `class` order change should in theory improve gzip compression. We should confirm this and if it doesn't I can revert the change.


jakearchibald:
I'm happy with this fwiw

XhmikosR:
I'm just not sure about the `class` attribute order change... Granted, it's a minor change, but could you verify it doesn't hurt gzip compression? I think it might does.

jakearchibald:
It'll depend on the kind of gzip github uses. In terms of brotli, there's no difference between the two. With zopfli, this PR is slightly smaller.

I'm happy to do with this PR.

PR #299, opened 2021-09-20
index.html: remove unneeded `div`s

XhmikosR:
@jakearchibald I think these are leftovers, but as usual, you should know better 😛

Non-whitespace diff: https://github.com/jakearchibald/svgomg/pull/299/files?w=1

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -109,53 +109,41 @@
           <div class="settings-scroller">
             <section class="global">
               <h1 class="settings-sub-heading">Global settings</h1>
-              <div>
-                <label class="setting-item-toggle">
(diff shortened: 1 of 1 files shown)


jakearchibald:
LGTM. Not sure why I had those divs in the first place. Might have been an error, or working around some old browser bug.

XhmikosR:
Have you tested Safari? I've tested Firefox and Chrome.

XhmikosR:
NVM, I just tested Safari 12, 14 and iOS on BrowserStack, it seems fine there too.

jakearchibald (approved):

PR #301, opened 2021-09-21
floating-action-button: rename `minor` parameter to `major`

XhmikosR:
Nitpicking mostly, but I figured since most buttons are small, this makes more sense.


src/js/page/ui/floating-action-button.js
```diff
@@ -3,7 +3,7 @@ import Ripple from './ripple';
 import Spinner from './spinner';
 
 export default class FloatingActionButton {
-  constructor({ title, href, iconSvg, classList, minor }) {
+  constructor({ title, href, iconSvg, classList, minor = true }) {
```
jakearchibald, line 6: I'm not keen on booleans that are true by default. Rename this `major`?

XhmikosR: Sure thing. It's just that we have a Sass file called minor-floating-button, but I plan to merge that.

jakearchibald (approved):

jakearchibald:
I've filed https://github.com/jakearchibald/svgomg/issues/334 for the remaining stuff here.

PR #303, opened 2021-09-21
Update pako to v2.0.4

XhmikosR:
While at it, I switched to their dist file. It seems they offer these files:

```
C:\Users\xmr\Desktop\svgomg\node_modules\pako\dist>dir /b *.js
pako.es5.js
pako.es5.min.js
pako.js
pako.min.js
pako_deflate.es5.js
pako_deflate.es5.min.js
pako_deflate.js
pako_deflate.min.js
pako_inflate.es5.js
pako_inflate.es5.min.js
pako_inflate.js
pako_inflate.min.js
```

I'm not sure if we should use something else. I tried `import { gzip } from 'pako';` but it seems it's including a lot more stuff obviously.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
         "gulp-sass": "^5.0.0",
         "gulp-sourcemaps": "^3.0.0",
         "nanoevents": "^6.0.1",
-        "pako": "^1.0.11",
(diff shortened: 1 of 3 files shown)


jakearchibald (approved):
LGTM

PR #308, opened 2021-09-22
Fix size mismatch when pasting an SVG

### LINKED ISSUE #140: Identical input/output incorrectly estimates output as larger
> 192 bytes → 193 bytes **100.52%**

Toggling Show Original in the Markup tab shows no change. Is it adding a trailing newline or something?

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="500" height="500" viewBox="0 0 500 500">
  <path d="M111 223L1 332l110 108v-81h194v-54H111v-82zm388-54L389 61v81H195v54h194v82l110-109z"/>
</svg>
```

XhmikosR:
Fixes #140

### DIFF
--- a/src/js/page/main-controller.js
+++ b/src/js/page/main-controller.js
@@ -133,7 +133,7 @@ export default class MainController {
   }
 
   _onGlobalPaste(event) {
-    const val = event.clipboardData.getData('text').trim();    
+    const val = event.clipboardData.getData('text');    
     if (!val.includes("</svg>")) {
       this._toastsUi.show("Pasted value not an SVG", {
         duration: 2000
--- a/src/js/page/svg-file.js
+++ b/src/js/page/svg-file.js
@@ -11,7 +11,7 @@ export default class SvgFile {
 
   async size({ compress }) {
     if (!compress) {
-      return this.text.length;
+      return this.text.trim().length;
     }
 
     if (!this._compressedSize) {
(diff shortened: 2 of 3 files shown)


jakearchibald:
Doesn't this mean the size becomes incorrect? If I save the file it'll be bigger than SVGOMG reports, since it'll have extra whitespace.

I think the real issue is that we trim in the input, so the input has been 'preprocessed'.

XhmikosR:
I believe it's a newline somewhere, but I'm not 100% sure. Because we already trim on paste.

XhmikosR:
@jakearchibald yeah, it's the trim on paste. I pushed another patch which removes trim from there.

XhmikosR:
BTW I really miss a Disable All / Enable All button 😛

XhmikosR:
I thought about this and maybe it makes more sense to go with the first patch. I mean, if the input has whitespace, the SVG size shouldn't include that whitespace. I do see your point though...

jakearchibald (approved):
I'm happy with this. It's SVGO's job to do the optimisation, and it'll trim as part of that. The benefit of having SVGO do it, is it reports that as part of the _output_, whereas previous to this PR we were applying optimisations as part of the input.

I think `.trim()` was used previously because it was convenient for the "is this SVG?" check, but that check changed.

PR #310, opened 2021-09-22
Specify `line-height` on the root element

### LINKED ISSUE #306: `results` jump on some browsers
Noticed this on Firefox on Android and on older iOS Safari. When the results are loaded there is a "jump".

See the attached video for a demo.

https://user-images.githubusercontent.com/349621/134286662-af85e1e2-e7fd-4be8-b050-3d88b0fe0ee8.mp4

XhmikosR:
Fixes #306

### DIFF
--- a/src/css/_global.scss
+++ b/src/css/_global.scss
@@ -7,7 +7,7 @@ html, body {
 
 html {
   font-family: roboto, sans-serif;
-  line-height: 1.3;
+  line-height: 1.25;
   -webkit-tap-highlight-color: rgba(0,0,0,0);
 
   // <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 2 2"><path d="M1 2V0h1v1H0v1z" fill-opacity=".05"/></svg>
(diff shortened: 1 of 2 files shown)


XhmikosR:
BTW this makes me wonder if we should set a root font-size and line-height. But that's for later

jakearchibald:
> BTW this makes me wonder if we should set a root font-size and line-height. But that's for later

Let's do the `line-height` bit in this PR. `line-height: 1.3` on the root element fixes this issue and looks good.

Good spot btw! Seems like it's the arrow char causing the issue. I guess it uses a different font with a different default line-height for that char.

XhmikosR:
Hmm, this changes a lot more cases, though. Are you OK with those changes?

I say we just go with the original patch which is the least intrusive 😛

XhmikosR:
1.25 seems to also work. Your call, I don't have strong feelings either. I'm used to using 1.5 and 1.25 for small text.

jakearchibald:
Yep, this works for me. Better to have line-height set everywhere. I realise it's a layout change across the site, but it looks fine to me. The difference is really minor.

jakearchibald:
(and it fixes the results issue)

PR #313, opened 2021-09-23
gulpfile: fix `SVGOMG_VERSION`

XhmikosR:


src/js/sw/index.js
```diff
@@ -1,8 +1,6 @@
-/* globals SVGOMG_VERSION:false */
-
 import {idbKeyval as storage} from '../utils/storage.js';
 
-const version = SVGOMG_VERSION;
```
TrySound, line 5: It's widely used practice to use global variable or process.env property. Your change is very implicit type casting.

XhmikosR: We don't need it to be a number since we use it as a String. And it's always replaced so it's not a global variable in the classic term.

TrySound: It's build step variable. I wouldn't get what's SVGOMG_VERSION string if tried to contribute the first time. With global variable it's clear there should be source of it.

PR #314, opened 2021-09-23
utils/storage.js: remove the iOS workaround

XhmikosR:
I tested things back to iOS 13 and it seems to work fine. iOS 12 doesn't work already on main.

Just so we are safe, someone should double check this. :)


jakearchibald (approved):
Double checked! LGTM

PR #315, opened 2021-09-23
touchDistance(): switch to `Math.hypot()`

XhmikosR:
Not sure about performance, but I figured that's the reason `Math.hypot()` exists.


jakearchibald (approved):
LGTM

PR #316, opened 2021-09-23
Add `type="button"` in buttons.

XhmikosR:
The default is `type="submit"`

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -89,7 +89,7 @@
       </div>
       <div class="toolbar">
         <div class="menu-toolbar-item">
-          <button class="unbutton menu-btn">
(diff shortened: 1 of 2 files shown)


TrySound:
Does it make sense outside of form element?

XhmikosR:
Yeah, it's good practice to explicitly set it.

jakearchibald:
Is there a benefit to this? It doesn't make it show up different to screen readers, for instance.

XhmikosR:
Now it's explicit what we expect the button to be. I mean, these are not forms. :)

And sure, if one's familiar enough with HTML will probably know this, but IMHO it's better to be explicit with the button type since when the type is missing the default is submit.

PR #317, opened 2021-09-23
Update README.md

XhmikosR:
@jakearchibald feel free to edit this. I felt the previous text was outdated. Basically the only true part is the "detailed options for each plugin" one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 ## Feature requests
 
-It's early days for this project, so it's missing important features such as code output view & detailed options for each plugin. [Check out the issues](https://github.com/jakearchibald/svgomg/issues) to see what's planned, or suggest ideas of your own!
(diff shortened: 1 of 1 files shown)


jakearchibald (approved):
Oh wow, didn't realise it was that out of date. Thanks!

XhmikosR:
BTW I feel like there's a duplication in the left menu:

* Contribute links to the repo
* About links to the README.md

Perhaps we should combine these two? Although I do feel that the menu is a little empty after that.

jakearchibald:
> Perhaps we should combine these two? Although I do feel that the menu is a little empty after that.

Good point. I'm happy with it looking emptier. The "about" link is pointless IMO. I'll remove it.

XhmikosR:
Haha https://github.com/jakearchibald/svgomg/commit/23d370781d24fb669eac626ce9dabb02c8833ffe

I'll open a PR!

PR #318, opened 2021-09-23
Remove underscore from public methods

XhmikosR:
I checked as good as I could and these are the ones I noticed; I was experimenting with mangling properties that start with an underscore.

I discovered I had to also [drop one loop in pan-zoom](https://github.com/jakearchibald/svgomg/commit/99664c8d16893fb334ffeccedda3eb22d5438dcb) and after these changes mangling seems to work fine. Should I push the loop drop here?

### DIFF
--- a/src/js/page/gzip.js
+++ b/src/js/page/gzip.js
@@ -6,7 +6,7 @@ class Gzip extends WorkerMessenger {
   }
 
   compress(svgData) {
-    return this._requestResponse({
+    return this.requestResponse({
       data: svgData
     });
   }
--- a/src/js/page/prism.js
+++ b/src/js/page/prism.js
@@ -6,7 +6,7 @@ export default class Prism extends WorkerMessenger {
   }
 
   highlight(svgData) {
-    return this._requestResponse({
+    return this.requestResponse({
       data: svgData
     });
   }
--- a/src/js/page/svgo.js
+++ b/src/js/page/svgo.js
@@ -8,7 +8,7 @@ export default class Svgo extends WorkerMessenger {
   }
 
   async wrapOriginal(svgText) {
-    const {width, height} = await this._requestResponse({
+    const {width, height} = await this.requestResponse({
       action: 'wrapOriginal',
       data: svgText
     });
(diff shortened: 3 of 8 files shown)


jakearchibald:
LGTM so far

XhmikosR:
@jakearchibald should I pick the for loop patch too? Otherwise you can push to this branch if you want, just note that there's no babel and I'm not sure if public fields support matches what you support here (which I don't know exactly BTW).

jakearchibald:
I'll do the public fields bit & push to this branch

XhmikosR:
Just wondering, from your analytics, how many are on Safari < 14? In general, only 2 latest versions are supported in this repo?

jakearchibald:
> Just wondering, from your analytics, how many are on Safari < 14? In general, only 2 latest versions are supported in this repo?

1.7% of users are on those old versions of Safari. Hmm ok, it's not worth breaking things for that many. I'll take a different approach.

Also, eslint's default parser doesn't support class fields, so it'd mean switching to babel's parser.

XhmikosR:
I think so far on main we support the latest 2 versions from all browsers. I tested Safari 13 and it worked on main (a few days ago).

Maybe we should still have a .browserlistrc file for reference even if it's unused? Or add a mention in README.md?

jakearchibald:
Yeah, it's fair to call it out as a policy in the readme

jakearchibald:
Let me know if you're happy with the removal of bind

XhmikosR:
Yup, the diff LGTM. I assume there's no impact in perf due to moving those in the constructor?

XhmikosR:
Diff-wise this is cleaner https://github.com/jakearchibald/svgomg/commit/2896ea7e603d194f238267d0d35e26d474b54415

Maybe we should go with this and I just add your comment?

jakearchibald:
> Diff-wise this is cleaner [2896ea7](https://github.com/jakearchibald/svgomg/commit/2896ea7e603d194f238267d0d35e26d474b54415)

I think we should optimise for code-readability over diff-readability.

The `.bind` pattern I was using becomes a readability problem, because it turns what looks like a prototype method into an instance methods. I'd rather it was obviously an instance method from its declaration. Public class fields are the best way to do this of course, but adding them in the constructor is a decent compromise until those Safari numbers drop.

XhmikosR:
Your call, I don't have strong feelings about this :)

jakearchibald (approved):
Let's go with this then

PR #322, opened 2021-09-24
Replace function expressions with callbacks

XhmikosR:
This was needed in the past due to the callback function's length, but now I think it's cleaner this way.


jakearchibald (approved):
Ah, this was needed in the past because the same function was passed to multiple listeners due to vendor prefixes, and that issue went away.

PR #323, opened 2021-09-24
transitionClassFunc(): use classList.toggle()

XhmikosR:
@jakearchibald Since we are either adding or removing the same class, isn't this the same?

### DIFF
--- a/src/js/page/utils.js
+++ b/src/js/page/utils.js
@@ -57,7 +57,7 @@ function transitionClassFunc({removeClass = false}={}) {
 
       requestAnimationFrame(() => {
         el.addEventListener('transitionend', listener);
-        el.classList[removeClass ? 'remove' : 'add'](className);
+        el.classList.toggle(className);
       });
     });
 


jakearchibald:
Doesn't this change the behaviour if `transitionClassFunc()` is called twice? It means parts of the function are operating on the parameter, whereas others are toggling, so there's a danger of getting out of sync.

XhmikosR:
I agree, but isn't this guaranteed to be called once? AFAICT that's the case, hence this patch.

jakearchibald:
`.toggle` ends up doing the right thing because there are early returns to prevent it doing the wrong thing. But doesn't that make the whole thing harder to read and more likely to break in future?

The code should express the intent.

XhmikosR:
TBH, it's the same to me, so the final call is yours. I, actually, find this cleaner.

Agreed about the code being from fragile, though.

jakearchibald:
I agree the current version is trying to be too smart. I'd be happy with:

```js
if (removeClass) {
  el.classList.remove(className);
} else {
  el.classList.add(className);
}
```

XhmikosR:
BTW I also have [another case](https://github.com/jakearchibald/svgomg/commit/2b36cc2a9c1c664142ee290526ef9244e8278726) in my branch, but I'm not sure I like it either. It does save some bytes, but it's probably less readable.

PR #324, opened 2021-09-24
bg-fill-button: move inline styles to CSS

XhmikosR:
IMHO having the styles in the CSS is always better.

Also, this fixes the wrong background after one toggled the dark background and went back to the default one because the default body background-color isn't transparent.

PS.

1. I plan to rename/move/cleanup _utils.scss later
2. I think we could also use `classList.toggle()` for the `active` class and skip the check. I have [a patch](https://github.com/jakearchibald/svgomg/commit/b322eb17656945c942d522c707745feaad24b9bc) on my other branch, let me know if you want me to push that here too.


jakearchibald:
> IMHO having the styles in the CSS is always better.

Changing class names on the root is expensive for the sake of changing one style. Although it's only a few ms in this case. Do you feel strongly about this? I'm not sure it's worth it.

XhmikosR:
I don't know how much overhead this might have, but it does fix another issue. Plus the utility could be reused.

If you still think we shouldn't go with toggling the class, I'll try another solution with `background-color: initial`.

jakearchibald:
Let's go with this, and just keep an eye on performance.

jakearchibald (approved):

jakearchibald (changes requested):
Actually, let's avoid using toggle here, and use add/remove to ensure everything stays in async.

It seems weird having one part of the code checking current state, then another calling toggle & hoping everything is in sync.

XhmikosR:
Should I add the class in the same `if` check then?

XhmikosR:
@XhmikosR Updated. Seems too verbose and I'm not sure if this can happen realistically. Personally, I'd just switch to `toggle` for `active` too, but it's your call.

XhmikosR:
@jakearchibald which of the patches do you prefer anyway? This is blocking me from switching to transitioning a pseudo selector :) (https://github.com/jakearchibald/svgomg/commit/314390646957983fe7eacfdfb026560d2bdbd4a3)

jakearchibald:
I'm happy with this PR

XhmikosR:
Yeah, but which of the 3 solutions? 😛

jakearchibald:
What are the three solutions? Can you give me commit or branch names so I can compare?

XhmikosR:
I have them as separate patches https://github.com/jakearchibald/svgomg/pull/324/commits

src/js/page/ui/bg-fill-button.js
```diff
@@ -16,17 +16,11 @@ export default class BgFillButton extends FloatingActionButton {
...
-    }
-  }
 
-  setColor(color) {
-    document.documentElement.style.backgroundColor = color;
+    const hasActiveClass = this.container.classList.contains('active');
+    const hasDarkClass = document.documentElement.classList.contains('bg-dark');
+
+    this.container.classList.toggle('active', !hasActiveClass);
+    document.documentElement.classList.toggle('bg-dark', !hasDarkClass);
```
jakearchibald, line 24:
```suggestion
  if (this.container.classList.contains('active')) {
    this.container.classList.remove('active');
    document.documentElement.classList.remove('bg-dark');
  } else {
    this.container.classList.add('active');
    document.documentElement.classList.add('bg-dark');
  }
```

The reason I prefer this style is it's clear what button state relates to which `<html>` background state. Unless i'm missing something in the other options, it isn't clear without knowing the _current_ state of `document.documentElement.classList`.

XhmikosR: Your suggested snippet relies on `active` for `bg-dark` too?

Isn't this what you prefer? https://github.com/jakearchibald/svgomg/pull/324/commits/5ed2ab80f6d9db1bad9b0fe2d2f5d8c3ef7b8c66

I kept the patches separately for this reason, so I can go back to whatever solution we want. :)

src/css/_utils.scss
```diff
@@ -30,3 +30,8 @@ $easeInOutSine: cubic-bezier(0.445, 0.05, 0.55, 0.95);
 $easeInOutExpo: cubic-bezier(1, 0, 0, 1);
 $easeInOutCirc: cubic-bezier(0.785, 0.135, 0.15, 0.86);
 $easeInOutBack: cubic-bezier(0.68, -0.55, 0.265, 1.55);
+
+// stylelint-disable declaration-no-important
+.bg-dark {
+  background-color: rgba(0, 0, 0, 0.7) !important;
```
jakearchibald, line 36: It doesn't seem like `!important` is necessary here?

XhmikosR: I added `!important` here because I consider this a utility where I expect it to always override the property.

jakearchibald:
```suggestion
  background-color: rgba(0, 0, 0, 0.7);
```

I'd rather avoid `!important` if possible. If we get to a situation where it's needed, we can tackle it then. But the cascade is working as expected right now.

XhmikosR: Fine by me, I'm used to it from the Bootstrap utils :)

PR #325, opened 2021-09-24
ripple: switch to a span

XhmikosR:
I was checking the serialized DOM for HTML validation errors, and came across this one. AFAICT it's safe.

The rest of the errors are harder to tackle and some are intentional like the missing `link` `rel` attribute which is intentional.

The `section` warning is invalid too since we add a heading when there's changelog but perhaps we could add the section only if there's a reason to show the changelog?

The `scrolling` error is also valid but I couldn't make it work properly...

val

**EDIT**: Just found another one:

```
Bad value 350.66 for attribute width on element iframe: Expected a digit but saw . instead
```

If browsers round when this happens, we could probably do it ourselves too with `Math.round`?

### DIFF
--- a/src/js/page/ui/ripple.js
+++ b/src/js/page/ui/ripple.js
@@ -2,7 +2,7 @@ import { strToEl } from '../utils';
 
 export default class Ripple {
   constructor() {
-    this.container = strToEl('<div class="ripple"></div>');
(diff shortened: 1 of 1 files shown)


jakearchibald (approved):
LGTM. For what it's worth, I'm not that bothered about validation errors unless it's causing a UX issue.

PR #326, opened 2021-09-24
svg-output.js: use `Math.round`

XhmikosR:
Floating point values aren't valid here.

~~I think that's the reason I was still seeing the iframe scrollbars on Firefox when I tried to remove `scrolling="no"`. Chrome didn't have this issue. I haven't tried Safari yet.~~

~~EDIT: the issue with scrollbars showing still happens on Firefox after zooming in on some cases :/ I'm out of ideas.~~
EDIT2: either way, we should probably use `Math.round` since floating point values are valid in width/height attributes anyway...


XhmikosR:
I dropped the `scrolling` patch, we should revisit that later.

jakearchibald (approved):

XhmikosR:
@jakearchibald another solution without using round on our side would be to use `style.width`/`style.height`. Maybe that would be better?

jakearchibald:
If that works, I'm happy with it

PR #328, opened 2021-09-26
main-controller: remove debugging code

XhmikosR:
It's basically the same as clicking the demo button.

Doesn't affect the production files, but it doesn't seem to offer a lot :)

### DIFF
--- a/src/js/page/main-controller.js
+++ b/src/js/page/main-controller.js
@@ -107,17 +107,6 @@ export default class MainController {
           duration: 3000
         });
       }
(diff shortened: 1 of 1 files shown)


jakearchibald:
I find this useful when making a change to UI that's only shown once an image is loaded. It avoids clicking that button each time. It also makes it easier to debug issues that happen with a particular SVG (by changing the SVG url)

XhmikosR:
All good, if you still use it :)

PR #330, opened 2021-09-26
svg-output: remove charset

XhmikosR:
This shouldn't be needed.

BTW @jakearchibald I think it's worth looking into doing something like https://github.com/tigt/mini-svg-data-uri. I tried to incorporate the package, but I don't think we need all of its features. We just need to normalize quotes and then just escape the characters that need to be escaped. The savings are a lot and it probably makes a difference with big SVG files.


jakearchibald:
> BTW @jakearchibald I think it's worth looking into doing something like https://github.com/tigt/mini-svg-data-uri.

I don't see the benefit. That library is useful for creating smaller inline SVGs, resulting in less HTML down the wire, but the dataurl here is never sent down the wire.

jakearchibald (approved):

Issue #280, opened 2021-09-15
Self host fonts


XhmikosR:

TrySound:
Why?

XhmikosR:
https://github.com/jakearchibald/svgomg/pull/250#issuecomment-916974688

mvoloskov:
Using system font stack would be a [great solution](https://mrmrs.cc/writing/webfonts/). And it looks good, too!

<img width="784" alt="image" src="https://user-images.githubusercontent.com/14220138/135074972-d157b41e-f535-4698-bedf-4e07e474c187.png">

This is the stack I applied:
```css
	font-family:
		-apple-system,
		BlinkMacSystemFont,
		"Segoe UI",
		Roboto,
		Oxygen,
		Ubuntu,
		Cantarell,
		"Fira Sans",
		"Droid Sans",
		"Helvetica Neue",
		sans-serif,
		"Apple Color Emoji",
		"Segoe UI Emoji",
		"Segoe UI Symbol";
```

XhmikosR:
Yup, agreed :) I'm going to make a PR later so that we test things out with a preview.

mvoloskov:
@XhmikosR I hope [my PR](https://github.com/jakearchibald/svgomg/pull/331) helps :)

XhmikosR:
@mvoloskov thanks for the PR!

That being said, there are a lot more leftovers. I will make a PR myself later after we get the OK from @jakearchibald :)

PR #331, opened 2021-09-28
Use system font stack

mvoloskov:
Instead of Roboto, I used system font stack for everything except code output. For that, I left Inconsolata in place as monospace system fonts aren't as good yet as sans-serifs.

This is [the reasoning](https://github.com/jakearchibald/svgomg/issues/280#issuecomment-929083176) behind this decision.

### DIFF
--- a/src/css/_global.scss
+++ b/src/css/_global.scss
@@ -6,7 +6,7 @@ html, body {
 }
 
 html {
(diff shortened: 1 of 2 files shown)


jakearchibald (approved):
Agreed. Using a web font for first-render text isn't worth it.

PR #333, opened 2021-09-29
Fix font CSS caching in SW

jakearchibald:


XhmikosR:
Although I think we can further tweak the font stack (we don't support all of those browsers), this is needed for things are they are.

jakearchibald:
Agreed. We can either bring the woff2 to the same origin, or find a way to avoid web fonts altogether (eg your branch)

XhmikosR:
I'm still working on it, I plan to make a PR to discuss it. But first I want to land #292 to avoid conflicts.

PR #339, opened 2021-10-01
Replace Google Fonts with system fonts.

XhmikosR:
@jakearchibald here it is. That being said I'm not 100% sure of all the changes:

1. I switched `sans-serif` to the primary font; not sure if you want this
2. I had to change some of the sizes, but I only tested Windows and Android
3. The material-slider font-size change is probably unrelated, but I feel like the current size is too small; even 0.85rem is small. I can revert this so that we tackle in a separate PR
4. Obviously, we can't get the exact same size, but open to suggestions :)

### DIFF
--- a/src/css/_global.scss
+++ b/src/css/_global.scss
@@ -7,8 +7,9 @@ body {
 }
 
 html {
-  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu, Cantarell, 'Fira Sans', 'Droid Sans', 'Helvetica Neue', sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol';
-  line-height: 1.25;
+  font-family: system-ui, -apple-system, 'Segoe UI', Roboto, 'Helvetica Neue', Arial, 'Noto Sans', 'Liberation Sans', sans-serif;
+  font-size: 1rem;
+  line-height: 1.3;
   -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
 
   // <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 2 2"><path d="M1 2V0h1v1H0v1z" fill-opacity=".05"/></svg>
@@ -21,8 +22,7 @@ code,
 kbd,
 samp,
 tt {
-  // stylelint-disable-next-line font-family-no-duplicate-names
-  font-family: monospace, monospace;
+  font-family: SFMono-Regular, Menlo, Monaco, Consolas, 'Liberation Mono', 'Courier New', monospace;
   font-size: 1em;
 }
 
--- a/src/index.html
+++ b/src/index.html
@@ -28,8 +28,6 @@
     </script>
     <script src="js/page.js" defer></script>
     <link rel="preload" as="style" href="all.css" onload="rel='stylesheet'">
-    <link rel="preload" crossorigin as="style" href="https://fonts.googleapis.com/css?family=Inconsolata" onload="rel='stylesheet'">
-    <link rel="preconnect" crossorigin href="https://fonts.gstatic.com">
     <link rel="manifest" href="manifest.json">
     <link rel="apple-touch-icon" href="imgs/icon.png">
     <link rel="icon" href="imgs/icon.png">
--- a/src/js/sw/index.js
+++ b/src/js/sw/index.js
@@ -5,8 +5,6 @@ import {idbKeyval as storage} from '../utils/storage.js';
 const version = SVGOMG_VERSION;
 const cachePrefix = 'svgomg-';
 const staticCacheName = `${cachePrefix}static-${version}`;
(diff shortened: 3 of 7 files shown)


XhmikosR:
I reverted the `code` font. I still see a FOUT when switching to the Code tab in some cases, but I have Service Workers disabled.

Let me know about the rest of the changes.

package.json
```diff
@@ -40,7 +40,7 @@
     "dev": "gulp dev",
     "lint": "npm run eslint && npm run stylelint",
     "eslint": "eslint --report-unused-disable-directives --ignore-path .gitignore .",
-    "stylelint": "stylelint src/css/ --rd --risd",
+    "stylelint": "stylelint src/css/**/*.scss --rd --risd",
```
XhmikosR, line 43: Are you sure this is needed? BTW when using wildcards here you need to double quote them.

jakearchibald: Originally I had the fonts in the CSS folder, but it wasn't the right place for them. I'll revert this change. Good catch!

XhmikosR:
@jakearchibald if we are going to self host the fonts, shouldn't we move the new CSS to a partial?

src/css/_fonts.scss
```diff
@@ -0,0 +1,26 @@
+@font-face {
```
XhmikosR, line 1: I wonder if we should also specify `font-display`?

src/js/sw/index.js
```diff
@@ -62,7 +62,7 @@ async function handleFontRequest(request) {
 addEventListener('fetch', event => {
   const url = new URL(event.request.url);
 
-  if (url.host == 'fonts.gstatic.com') {
+  if (url.pathname.endsWith('.woff2')) {
```
XhmikosR, line 65: Just wondering, does it make sense to keep the separate font cache?

gulpfile.js
```diff
@@ -97,7 +97,7 @@ const allJs = gulp.parallel(
 
 function copy() {
   return gulp.src([
-    'src/{.well-known,imgs,test-svgs}/**',
+    'src/{.well-known,imgs,test-svgs,fonts}/**',
```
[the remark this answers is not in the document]
jakearchibald: Yep, good catch

src/js/sw/index.js
jakearchibald: It does for the latin-ext and vietnamese font, else they'll be lost in version bumps

XhmikosR: Do we even need to treat those differently? I mean, we only have latin characters :)

XhmikosR: Ah, true, thanks!

src/css/_global.scss
```diff
@@ -21,8 +22,7 @@ code,
 kbd,
 samp,
 tt {
-  // stylelint-disable-next-line font-family-no-duplicate-names
-  font-family: monospace, monospace;
+  font-family: SFMono-Regular, Menlo, Monaco, Consolas, 'Liberation Mono', 'Courier New', monospace;
```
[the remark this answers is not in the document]
jakearchibald: Yeah, it feels better than the default `monospace` in case monospace is needed outside of the code view.

I could put it in a custom property to avoid the repetition if you have strong feels about it.

src/index.html
```diff
@@ -28,8 +28,7 @@
     </script>
     <script src="js/page.js" defer></script>
     <link rel="preload" as="style" href="all.css" onload="rel='stylesheet'">
-    <link rel="preload" crossorigin as="style" href="https://fonts.googleapis.com/css?family=Inconsolata" onload="rel='stylesheet'">
-    <link rel="preconnect" crossorigin href="https://fonts.gstatic.com">
+    <link rel="preload" crossorigin as="font" href="fonts/code-latin.woff2">
```
XhmikosR, line 31: Could we maybe not preload the code font at all? I mean, the initial page doesn't need it.

Just wondering, again, I swear it is my last comment on this PR :)

src/css/_global.scss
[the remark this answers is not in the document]
XhmikosR: Nah, all good, I figured that these selectors are for normalization.

src/index.html
jakearchibald: It's good to preload it before it's needed, to reduce 'block' delay. I guess it could be preloaded once an SVG is loaded, but it feels like a bit of micro-optimisation. It's only 11.5kB, it isn't render blocking, and it isn't getting in the way of more important resources.

jakearchibald (approved):

XhmikosR:
@jakearchibald I thought about this and I don't think we should preload the code font after all. I mean, the same way we don't load the JS files needed for syntax highlighting.

How about we stop preloading the code font and see how it goes? If we notice any serious issues we can preload it again.

jakearchibald:
What's the downside to preloading the font?

The downside to not preloading the font is FOUT.

XhmikosR:
Also, in most cases, the font should be already fetched by the SW, right? So, there shouldn't be any FOUT for the majority of users.

PR #340, opened 2021-10-01
Deduplicate floating button code

XhmikosR:
Rename minor to major and inverse the logic.

Depends on #335 and other stuff, but I will rebase so that we land this separately.

Notes:

* `fill` is not needed for the minor button, but I left it there for consistency
* I removed `FloatingActionButton`'s `classList` parameter since it's not used


src/css/components/_floating-action-button.scss
```diff
@@ -1,20 +1,22 @@
...
-  background-color: #00bcd4;
+  width: 40px;
+  height: 40px;
+  border-radius: 20px;
+  background-color: #fff;
   box-shadow: 0 2px 3px rgba(0, 0, 0, 0.25);
   position: relative;
+  cursor: pointer;
+  overflow: hidden;
+  z-index: 0; // Is this really needed?
```
XhmikosR, line 10: Not sure if this is needed...

Issue #311, opened 2021-09-22
PageSpeed Insights/Lighthouse tasks


XhmikosR:
https://lighthouse-dot-webdotdevsite.appspot.com//lh/html?url=https%3A%2F%2Fjakearchibald.github.io%2Fsvgomg%2F

1. for a11y I've made #304 and I'll make some more patches later;
2. for the font we have #280
3. this leaves us these 2:
  *  Does not use passive listeners to improve scrolling performance
  *  Document does not have a meta description

Now, PageSpeed shows a different picture on mobile, not sure if they use a different emulation config: https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fjakearchibald.github.io%2Fsvgomg%2F

PR #343, opened 2021-10-03
performance: removes lighthouse warning about passive scroll

luckened:
This fixes lighthouse warning about using `passive listeners to improve scrolling performance`:

image

ive tested in home, let me know if i missed something

one of #311 tasks

please consider tagging it as 'hacktoberfest-accepted' 🥳

### DIFF
--- a/src/js/page/ui/settings.js
+++ b/src/js/page/ui/settings.js
@@ -33,7 +33,7 @@ export default class Settings {
       this._scroller = document.querySelector('.settings-scroller');
 
       this.container.addEventListener('input', e => this._onChange(e));
-      this._scroller.addEventListener('wheel', e => this._onMouseWheel(e));
+      this._scroller.addEventListener('wheel', e => this._onMouseWheel(e), { passive: true });
       this._resetBtn.addEventListener('click', e => this._onReset(e));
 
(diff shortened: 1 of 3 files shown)


XhmikosR:
Thanks for the PR. That being said, this isn't right. Pan and zoom needs `preventDefault()` so we can't use passive there.

src/js/page/ui/settings.js
```diff
@@ -33,7 +33,7 @@ export default class Settings {
       this._scroller = document.querySelector('.settings-scroller');
 
       this.container.addEventListener('input', e => this._onChange(e));
-      this._scroller.addEventListener('wheel', e => this._onMouseWheel(e));
+      this._scroller.addEventListener('wheel', e => this._onMouseWheel(e), { passive: true });
```
XhmikosR, line 36: This is also wrong as the code is since `_onMouseWheel` is using `preventDefault` too.

XhmikosR: But perhaps the Firefox workaround is outdated. We probably need to revisit the code.

XhmikosR:
Thanks for the PR, we are going to revisit these later ourselves :)

PR #345, opened 2021-10-13
Add social tags

jakearchibald:
Adding tags for social sharing


jakearchibald:
@XhmikosR it seems like all these tags are required.

XhmikosR:
Agreed, but if you are not in a hurry, I can check this closer a little later :)

PR #347, opened 2021-10-14
Assorted lint tweaks

XhmikosR:
@jakearchibald I've kept the patches separate but let me know if you want me to squash some of them.

Nothing is set in stone, open to suggestions, as usual :)

1. Not 100% sure about the params destructuring patch since it's not enforced and it results to slightly bigger files (not sure about any perf impact, probably negligible)
2. In the autofix patch I've added some comments that will show up when linting too. It'd be a good chance to revisit any of them (for example, can't we include the test SVG file in the SW so that it's available offline too?)
3. Reviewing is easier with `?w=1` in the URL

### DIFF
--- a/src/js/page/main-controller.js
+++ b/src/js/page/main-controller.js
@@ -21,23 +21,23 @@ const svgo = new Svgo();
 
 export default class MainController {
   constructor() {
-    this._container = null;
-
     // ui components
     this._mainUi = null;
     this._outputUi = new Output();
     this._downloadButtonUi = new DownloadButton();
     this._copyButtonUi = new CopyButton();
-    this._bgFillUi = new BgFillButton();
     this._resultsUi = new Results();
     this._settingsUi = new Settings();
     this._mainMenuUi = new MainMenu();
     this._toastsUi = new Toasts();
-    this._dropUi = new FileDrop();
-    this._preloaderUi = new Preloader();
-    this._changelogUi = new Changelog(self.version);
+
+    const bgFillUi = new BgFillButton();
+    const dropUi = new FileDrop();
+    const preloaderUi = new Preloader();
+    const changelogUi = new Changelog(self.version);
+    // _resultsContainerUi is unused
     this._resultsContainerUi = new ResultsContainer(this._resultsUi);
-    this._viewTogglerUi = new ViewToggler();
+    const viewTogglerUi = new ViewToggler();
 
     // ui events
     this._settingsUi.emitter.on('change', () => this._onSettingsChange());
@@ -47,13 +47,11 @@ export default class MainController {
     this._mainMenuUi.emitter.on('svgDataLoad', (event) =>
       this._onInputChange(event),
     );
-    this._dropUi.emitter.on('svgDataLoad', (event) =>
-      this._onInputChange(event),
-    );
+    dropUi.emitter.on('svgDataLoad', (event) => this._onInputChange(event));
     this._mainMenuUi.emitter.on('error', ({ error }) =>
       this._handleError(error),
     );
-    this._viewTogglerUi.emitter.on('change', (event) =>
+    viewTogglerUi.emitter.on('change', (event) =>
       this._outputUi.set(event.value),
     );
     window.addEventListener('keydown', (event) => this._onGlobalKeyDown(event));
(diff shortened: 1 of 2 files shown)


src/js/svgo-worker/index.js
```diff
@@ -10,10 +10,13 @@ const createDimensionsExtractor = () => {
         element: {
           enter: ({ name, attributes }, { type }) => {
             if (name === 'svg' && type === 'root') {
-              if (attributes.width != null && attributes.height !== null) {
+              if (
+                attributes.width !== undefined &&
```
XhmikosR, line 14: This one is more verbose, happy to go with `!= null` for both checks; previously one of them used the strict comparison which might have been wrong

src/js/page/ui/view-toggler.js
```diff
@@ -16,13 +16,11 @@ export default class ViewToggler {
       // stop browsers remembering previous form state
       this.container.output[0].checked = true;
 
-      this.container.addEventListener('change', () => this._onChange());
```
XhmikosR, line 19: I can revert this if you prefer and I agree this might be consistent with the rest of the codebase. I just noticed that the minifier doesn't inline the functions used once. I didn't check any other places, just happened to notice this one because the file is small.

src/js/page/main-controller.js
```diff
@@ -58,64 +68,66 @@ export default class MainController {
...
-        });
+        this._toastsUi.show('Ready now!', { duration: 3000 });
       }
 
       // for testing
       // eslint-disable-next-line no-constant-condition
       if (false) {
         (async () => {
-          this._onInputChange({
-            data: await fetch('test-svgs/car-lite.svg').then(r => r.text()),
```
jakearchibald, line 116: I'm not against this change, but why? I thought it was to avoid putting await within another expression, but another change `(await toast.answer)` does exactly that.

src/js/page/results-cache.js
```diff
@@ -1,3 +1,5 @@
+// TODO: switch to Map/Set?
```
jakearchibald, line 1: Yeah, a map would work well here, and the code would likely read better. Doesn't need to be done for this PR tho

src/js/page/ui/bg-fill-button.js
```diff
@@ -21,8 +21,7 @@ export default class BgFillButton extends FloatingActionButton {
     if (this.container.classList.contains('active')) {
       this.container.classList.remove('active');
       this.setColor('transparent');
-    }
-    else {
```
jakearchibald, line 25: wow I forgot I used to do this

src/js/page/ui/floating-action-button.js
```diff
@@ -1,30 +1,25 @@
...
-    if (href) {
-      this.container.href = href;
-    }
-    if (title) {
-      this.container.setAttribute('title', title);
-    }
-    if (major) {
-      classes.push('major-floating-action-button');
-    }
+    if (href) this.container.href = href;
```
jakearchibald, line 15: I used to be against single line ifs, but not anymore. This is so much easier to read 😄

src/js/page/ui/pan-zoom.js
```diff
@@ -1,37 +1,34 @@
-function getXY(obj) {
+const getXY = (obj) => {
```
jakearchibald, line 1: Not against this change, but what's the reasoning?

src/js/page/ui/settings.js
```diff
@@ -109,35 +113,28 @@ export default class Settings {
...
 
-    this._pluginInputs.forEach(inputEl => {
+    for (const inputEl of this._pluginInputs) {
       fingerprint.push(Number(inputEl.checked));
       output.plugins[inputEl.name] = inputEl.checked;
-    });
+    }
 
-    output.fingerprint = fingerprint.join();
+    output.fingerprint = fingerprint.join(',');
```
jakearchibald, line 137: This isn't quite right. The idea is:

- Inputs other than `gzip` and `original` are used to form the fingerprint.
- All inputs values are gathered into `output`.

Right now it's including `gzip` and `original` in the fingerprint. The first `if` is never true.

src/js/page/ui/toasts.js
```diff
@@ -4,8 +4,9 @@ class Toast {
   constructor(message, duration, buttons) {
     this.container = strToEl('<div class="toast"><div class="toast-content"></div></div>');
 
-    this._content = this.container.querySelector('.toast-content');
-    this._content.textContent = message;
+    const content = this.container.querySelector('.toast-content');
```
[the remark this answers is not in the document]
jakearchibald: Good catch. Yeah, no point making an instance property if it's only needed in one function.

src/js/page/ui/pan-zoom.js
jakearchibald: I don't mind top level functions (I think brotli does a good job here), but eh, happy to land this as is to avoid more churn.

src/js/page/ui/settings.js
```diff
@@ -46,7 +46,7 @@ export default class Settings {
     // Prevents bounce effect on desktop.
     // Firefox uses DELTA_LINE on a mouse wheel, ignore it
     // 0 is "pixels"
-    if (!event.deltaMode) {
+    if (event.deltaMode === 0) {
```
XhmikosR, line 57: Just to confirm, I don't suppose this can be undefined or null? My understanding is that this is a Firefox-specific workaround where `deltaMode` is `0`.

jakearchibald: It's always a number, yeah

Issue #307, opened 2021-09-22
settings are not completely aligned


XhmikosR:
<img src="https://user-images.githubusercontent.com/349621/134287437-fc3d636f-8e15-415f-bc1f-e91826a48d59.png" alt="settings" width="300">

PR #351, opened 2021-10-18
BUG #307 - settings are not completely aligned

Graxi:
- Fix the alignment in settings section
image

### DIFF
--- a/src/css/components/_settings.scss
+++ b/src/css/components/_settings.scss
@@ -37,6 +37,11 @@
       box-shadow: 0 1px 3px rgba(0, 0, 0, 0.5), inset 0 0 0 2px #3f51b5;
     }
   }
+
+  .material-switch + span {
+    display: inline-block;
+    width: calc(100% - 56px);
+  }
 }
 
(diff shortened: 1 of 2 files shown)


Graxi:
> This change feels a bit hacky to me, but it's not your fault, it's flexbox's fault. I avoid flexbox these days if I can, it just doesn't behave how I expect in cases like this, or it requires extra markup to do simple things 😄
> 
> Could you fix this by making each `setting-item-toggle` a two column grid? Also, make the gap between the switch and text a `gap` rather than a margin on the `material-switch`.
> 
> You should be able to revert the markup change to `index.html`.

Hi I have updated the PR and made `setting-item-toggle` two column structure. Would you like to review again? Let me know if further changes are needed please. I am not sure if I make it correct since it's my first commit to open source;)

jakearchibald:
@XhmikosR

> Shouldn't we wrap all items, though? It seems a little fragile to only do this for specific options only.

I'm not suggesting doing this for specific options, I'm suggesting doing it for all of them.

PR #353, opened 2021-10-18
Improve Error messages display

XhmikosR:
| Main | First patch | Second patch|
|------|-------------|---------------|
| | |  |

The button's height should probably be limited. Feel free to push any improvements and let me know which approach you prefer.

### DIFF
--- a/src/js/page/ui/toasts.js
+++ b/src/js/page/ui/toasts.js
@@ -1,16 +1,20 @@
 import { strToEl, transitionToClass } from '../utils.js';
 
 class Toast {
-  constructor(message, duration, buttons) {
+  constructor(message, duration, buttons, isError) {
     this.container = strToEl(
       '<div class="toast"><div class="toast-content"></div></div>',
     );
-
     const content = this.container.querySelector('.toast-content');
-    content.textContent = message;
-
     this._answerResolve = null;
     this._hideTimeout = null;
+
+    if (isError) {
+      content.insertAdjacentHTML('afterbegin', '<pre><code></code></pre>');
+      content.querySelector('code').textContent = message;
+    } else {
+      content.textContent = message;
+    }
 
     this.answer = new Promise((resolve) => {
       this._answerResolve = resolve;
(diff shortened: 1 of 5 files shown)


src/js/page/ui/toasts.js
```diff
@@ -1,17 +1,21 @@
...
     );
-
     const content = this.container.querySelector('.toast-content');
-    content.textContent = message;
-
     this._answerResolve = null;
     this._hideTimeout = null;
 
+    if (isError) {
+      content.insertAdjacentHTML('afterbegin', '<pre><code></code></pre>');
```
XhmikosR, line 13: If there's a simpler way to do this, let me know.

jakearchibald: I'm happy with this approach

XhmikosR:
So you prefer the second patch, right?

src/js/page/ui/toasts.js
```diff
@@ -1,17 +1,21 @@
 import { strToEl, transitionToClass } from '../utils.js';
 
 class Toast {
-  constructor(message, duration, buttons) {
+  constructor(message, duration, buttons, isError) {
```
jakearchibald, line 4: Move `isError` to an option to avoid the boolean trap?

XhmikosR: Not sure I follow you here... BTW we don't call `Toast` anywhere else except from the `Toasts` Class.

XhmikosR: At some point I played with moving the Toast class to a separate file, but I went back and forth because it's only used here. If you think that would be better let me know.

jakearchibald:
> So you prefer the second patch, right?

Yes

jakearchibald:
Before I do a full review, is the "second patch" the one represented by this PR?

XhmikosR:
Yup. It's the one with the monospaced font.

jakearchibald (approved):
LGTM!

PR #354, opened 2021-10-18
pan-zoom.js: remove Firefox workaround

XhmikosR:
Not sure what exactly this was supposed to fix/work around, but it might not be needed anymore?

### DIFF
--- a/src/js/page/ui/pan-zoom.js
+++ b/src/js/page/ui/pan-zoom.js
@@ -113,16 +113,9 @@ export default class PanZoom {
     event.preventDefault();
 
     const { left, top } = this._target.getBoundingClientRect();
-    let delta = event.deltaY;
-
-    // 1 is "lines", 0 is "pixels"
-    // Firefox uses "lines" when mouse is connected
-    if (event.deltaMode === 1) {
-      delta *= 15;
-    }
 
     // stop mouse wheel producing huge values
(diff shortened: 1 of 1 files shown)


jakearchibald (commented):
It does seem like Firefox is now using a `deltaMode` of pixels rather than lines, but that doesn't stop 'lines' being a thing. Is it worth keeping the workaround here for other possible cases of this, given that it's standard behaviour?

jakearchibald:
It seems like this is still an issue pretty recently https://github.com/pmndrs/use-gesture/issues/236. I can't recreate it locally, but maybe it depends on the type of mouse, or maybe Firefox changed things more recently, but I can't find anything in their bug tracker about it.

XhmikosR:
We can put this on hold if you think it's better. I couldn't replicate the issue personally using the latest Firefox.

jakearchibald:
I couldn't replicate it either, but unlike https://github.com/jakearchibald/svgomg/pull/355, the thing being removed in this PR addresses a feature, not a bug. That feature still exists. https://github.com/pmndrs/use-gesture/issues/236 suggests it's recently used in the wild.

Yeah, I don't think we should ship this change.

XhmikosR:
Marked the PR as draft.

Perhaps we should then change the comment? I'm fine with dropping this patch, I was just going through the codebase for stuff that might be outdated :)

PR #355, opened 2021-10-18
settings.js: remove workaround

XhmikosR:
Same as #354 but for settings, let me know if it's still needed.

We need to test Safari on desktop and iOS too.


jakearchibald:
Tested this across desktop and mobile browsers, it doesn't seem to be a problem anymore. Some browsers used to trigger a bounce effect in the main window when the settings scroller reached its ends.

PR #356, opened 2021-10-18
Add a separate apple-touch-icon.png

XhmikosR:
Firefox is also requesting this icon BTW. AFAICT it doesn't need to be bigger than 180x180 nowadays.

Unsure if it should be added in the SW.

BTW I'm pretty sure there is a reason you are using one icon for favicon, apple touch icon, social etc (no extra requests is my guess), but can't we do better? Especially in the favicon case, a 48x48 PNG file is a lot smaller. Not sure what's the best solution nowadays; I still use different icons, personally.


jakearchibald:
What's the benefit here? If the only benefit is shaving 1kB off a non-blocking resource, then I don't think it's worth having to maintain multiple copies of the asset.

XhmikosR:
Well, yeah. I personally dislike useless bytes even if it's handier to just use one icon for everything. And give that apple-touch-icon's specs have been the same for years, it might make sense to use a dedicated 180x180 icon.

I see your point, though, even if I don't agree :)

BTW, I also planned to do the same for the other icons because the difference is big IMHO: https://github.com/jakearchibald/svgomg/commit/5b76e5dbe71e938f7594a4a3f8b3872b5ba783c6. I just don't know if it's worth doing it for the Android icons; the favicon case seems make a lot of sense, though. We could also switch to an SVG favicon later and just keep the PNG one as a fallback?

jakearchibald:
I don't think 1kB is big when it comes to a non-blocking resource. If we were talking about 20kB, sure maybe it'd be worth it.

I feel the same when it comes to providing multiple image formats fwiw. I wouldn't do PNG+WebP if the WebP was only 1kB smaller, since it's not worth the complexity.

PR #358, opened 2021-10-19
.output-switcher: stop setting the filter property

XhmikosR:
Same reasons as #357.

We can keep it if you still think it's worth having it. I personally have trouble noticing the difference, but maybe it's just me.


jakearchibald (approved):
Yeah, I don't mind losing this effect

PR #359, opened 2021-10-19
Remove title from inline SVGs

XhmikosR:
The parent element already has the title set. Also, add `aria-hidden="true"` in the SVGs.

This should still be OK, but please confirm.

/CC @patrickhlauke if you have a little time to confirm :)


XhmikosR:
Perfect, thanks! I plan to tackle more a11y issues but I'm going step by step because it's an old codebase :)

jakearchibald (approved):
LGTM and thanks for reviewing @patrickhlauke!

Issue #304, opened 2021-09-22
Improve keyboard navigation


XhmikosR:
Currently, keyboard navigation is broken in many places.

As a first step, I think we should switch to `button`s when possible (I have a WIP patch), and then see how we can solve the remaining issues with material tabs and the floating buttons getting focused when they are not being shown.

I mark this as a bug because IMHO this is an important accessibility issue.

PR #363, opened 2021-10-19
Main menu: improve accessibility

XhmikosR:
@patrickhlauke: can you please check this out too?

BTW apart from the keyboard navigation which is [broken](#304) due to not properly hiding some elements, if you notice any other issues let me know. I do have another a11y patch in my `xmr-dev` branch (https://github.com/jakearchibald/svgomg/commit/8694a247f80b25e8ddb8d27ca319a675245415ef), but I need to probably split it because it might not be 100% correct, so we could use your help to at least identify any other issues :)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -43,7 +43,7 @@
     <div class="app-output">
       <div class="main-menu">
         <div class="overlay"></div>
-        <nav class="menu">
+        <nav id="menu" class="menu" aria-expanded="true">
           <div class="app-title">
             {% include "partials/icons/logo.svg" %}
             <div class="title-text">
@@ -86,7 +86,7 @@
       </div>
       <div class="toolbar">
         <div class="menu-toolbar-item">
-          <button class="unbutton menu-btn" type="button">
+          <button class="unbutton menu-btn" type="button" title="Menu" aria-controls="menu">
             {% include "partials/icons/menu.svg" %}
           </button>
         </div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -86,7 +86,7 @@ <h1>SVGOMG</h1>
       </div>
       <div class="toolbar">
         <div class="menu-toolbar-item">
-          <button class="unbutton menu-btn" type="button">
+          <button class="unbutton menu-btn" type="button" title="Menu" aria-controls="menu">
```
patrickhlauke, line 89: `aria-controls` is mostly pointless, as it's hardly supported. https://a11ysupport.io/tech/aria/aria-controls_attribute

src/index.html
```diff
@@ -43,7 +43,7 @@
     <div class="app-output">
       <div class="main-menu">
         <div class="overlay"></div>
-        <nav class="menu">
+        <nav id="menu" class="menu" aria-expanded="true">
```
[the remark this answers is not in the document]
patrickhlauke: more fundamentally, as the menu acts like a modal (or rather, it should, once fixed properly), you can save yourself the expanded/collapsed state on the button altogether. once the menu button is activated, the slide-in modal menu appears, focus should be moved to it, and kept inside it. underlying document should be hidden from assistive technologies and keyboard users. so you'd not get back to that button trigger anyway to hear that it's now "expanded" the menu. once menu slide-in closes, move focus back to the menu trigger button

src/partials/icons/menu.svg
```diff
@@ -1 +1 @@
-<svg viewBox="0 0 24 24"><title>Menu</title><path d="M3 18h18v-2H3v2zm0-5h18v-2H3v2zm0-7v2h18V6H3z"/></svg>
+<svg viewBox="0 0 24 24" aria-hidden="true"><path d="M3 18h18v-2H3v2zm0-5h18v-2H3v2zm0-7v2h18V6H3z"/></svg>
```
patrickhlauke, line 1: from all the changes in this PR, this is possibly the only one i'd keep (and for the rest consider looking more at how the bootstrap offcanvas component now operates)

XhmikosR:
Thanks for the comments @patrickhlauke! The focus stuff need to be handled another time for everything :)

PR #365, opened 2021-11-04
settings: move fingerprint to the output variable

XhmikosR:


src/js/page/ui/settings.js
```diff
@@ -121,11 +121,11 @@ export default class Settings {
     }
 
     for (const inputEl of this._pluginInputs) {
-      fingerprint.push(Number(inputEl.checked));
+      output.fingerprint.push(Number(inputEl.checked));
       output.plugins[inputEl.name] = inputEl.checked;
     }
 
-    output.fingerprint = fingerprint.join(',');
+    output.fingerprint = output.fingerprint.join(',');
```
jakearchibald, line 128: What's the benefit of doing it this way? It feels like this makes the code harder to read, as the type is changed from its initialisation.

XhmikosR: TBH, it's the same in my eyes, but feel free to drop the PR, I don't have strong feelings about it.

PR #366, opened 2021-11-04
Use `buttons` when possible

XhmikosR:
@patrickhlauke: can you check this out and #367/#368 please? ❤


jakearchibald:
I'm happy for this to be landed as is. The only reason they were divs rather than buttons is due to a Firefox bug where particular layouts within the button failed. They appear to be fixed.

PR #367, opened 2021-11-04
Add `aria-label` in the load file input

XhmikosR:


jakearchibald:
I don't think this is necessary. The element has `display: none` so it doesn't appear in the a11y tree.

XhmikosR:
I see it being flagged hence this patch. Feel free to close it if you are sure it's fine :)

jakearchibald:
> I see it being flagged hence this patch

What's flagging it?

XhmikosR:
https://squizlabs.github.io/HTML_CodeSniffer/ not sure about axe

jakearchibald:
I think it's misleading to do accessibility work on something that never appears in the AT.

PR #369, opened 2021-11-04
bg-fill-button: use a color when the button is active

XhmikosR:


jakearchibald (approved):
LGTM

jakearchibald:
@XhmikosR I've made a tweak so the ripple still works. Happy with this?

XhmikosR:
Yup, all good, I wasn't sure about the ripple effect so I left it as it was (the dummy way).

PR #370, opened 2021-11-04
svg-output: remove pointer events workaround

XhmikosR:
@jakearchibald AFAICT this is the whole workaround code, but you might have a better recollection. Could you verify the issue doesn't happen anymore?


jakearchibald (approved):
Confirmed. This isn't an issue anymore.

Issue #184, opened 2020-05-12
[Feature Request] Copy settings as CLI string


ptim:
I love previewing SVGO settings in SVGOMG, thanks!

It would be fantastic to have a button at near the config settings "Copy settings as CLI", like:

`npx svgo test.svg --precision=n  --config="<ESCAPED_JSON_STRING>"`

ericbf:
I'd like to see this too!

XhmikosR:
@jakearchibald this seems duplicate of #11

ptim:
I missed the previous ticket - I'll close this one and continue to follow over at #11 👍

WebMechanic:
I'd love to see **anything like this**, too!

There are now 3 separate request **since 2015** to implement this feature in one way or another #11, #50 and this one #184.

Issue #50, opened 2016-06-28
Export chosen settings as a command line to run command version of svgo


theghostbel:

Haroenv:
Would love to see this happen!

ericcornelissen:
Just for administration, isn't this a duplicate of #11?

Certainly would be interested in this as well tho 😄

harold:
+1

XhmikosR:
Duplicate of #11

Issue #11, opened 2015-01-27
Show cmd line equivalent


jakearchibald:
I'm not interested in a batch mode, but a view that shows the equivalent cmd will enable this

Dogbombs:
Absolutely.  Anyone with a reasonable amount of command-line Fu should be able to take the command shown and put it in a batch.  Thanks!

nobodypb:
+1
Also a svgo settings file (.svgo.yml) with the current settings for download would be nice

Tigt:
Might help when debugging as well, for when those rogue SVGs that SVGO chokes on get input.

bkuri:
+1
PS, thanks for the awesome tool.

colinmcglynn:
+1
I would love to be able to use this tool to test out some different settings and then use the exported settings to run a batch on the command line.

motleydev:
+1

harold:
+1

WebMechanic:
There are now 3 separate request **since 2015** to implement this feature in one way or another: this one #11, #50 and #184 (closed in favour of this).

PR #383, opened 2022-04-18
Add ability to download resulting config

epodivilov:
Hi.

I and others (according to requests https://github.com/jakearchibald/svgomg/issues/11, https://github.com/jakearchibald/svgomg/issues/50 and https://github.com/jakearchibald/svgomg/issues/184), miss the possibility to apply SVGO configuration to multiple files. As I understood it, batch processing support is not expected, that's why I suggest an intermediate solution - a possibility to download the resulting config and then apply it using the cli interface of SVGO.

Unfortunately, the latest versions of SVGO don't support sending the config as text, that's why you'll have to create the config file as JS.

```bash
# Example of usage
npx svgo --config="./path/to/svgo.config.js" -f ./input -o ./output
```


src/js/page/ui/settings.js
```diff
@@ -83,6 +89,47 @@ export default class Settings {
...
   }
 
+  _onExport() {
+    this._exportRipple.animate();
+
+    const { fingerprint, multipass, pretty, ...settings } = this.getSettings();
+    const floatPrecision = Number(settings.floatPrecision);
+    const plugins = [];
+
+    for (const [name, active] of Object.entries(settings.plugins)) {
```
XhmikosR, line 99: Most of this block seems to be duplicated in svgo-worker. Doesn't seem good from a maintenance perspective.

epodivilov: I decided to copy the code first. I can rework it a bit and take out the repeating code.

XhmikosR: I think that would help, yeah. Because this kind of duplication can be bad in the future.

Maybe we should just have a function to return the config with plugins somewhere in a common place.

@jakearchibald any suggestions?

src/js/utils/download.js
```diff
@@ -0,0 +1,15 @@
+export function downloadSvgoConfig(filename, text) {
+  const element = document.createElement('a');
```
XhmikosR, line 2: I'm wondering, isn't there a cleaner way without adding the extra `a` tag?

epodivilov: Rewrote the code so that the link to the configuration was always on the page and was updated when the configuration was changed

src/js/page/ui/settings.js
epodivilov: Moved the code for getting plugins into a separate utility

XhmikosR:
hey, there. Thanks for continuing your work on this.

Can you please make sure there are no lint issues and the the export actually works?

src/js/utils/download.js
```diff
@@ -0,0 +1,3 @@
+export function createFileURL(data, type) {
```
XhmikosR, line 1: This doesn't need to be a new file AFACT since it;s only used once.

XhmikosR:
This doesn't seem to work, no lint errors, no console errors but the config is not downloaded, at lean of Firefox that I tried.

epodivilov:
> This doesn't seem to work, no lint errors, no console errors but the config is not downloaded, at lean of Firefox that I tried.

What version of Firefox do you have? 

I have tested with the following versions on my system (MacOS Ventura) 
- Chrome 107.0.5304.87
- Firefox 106.0.4 
- Safari 16.1

I tested the export button on localhost and on the netlify build

XhmikosR:
Maybe you are missing a `click` event?

epodivilov:
> Maybe you are missing a `click` event?

But it's a link, it doesn't need click event handler. 
Maybe you have an old variant with button cached?

XhmikosR:
Yup, that was it.

Now, the only issue I see is ripple not firing when clicking the button.

src/js/utils/settings.js
```diff
@@ -0,0 +1,28 @@
+export function collectPlugins(settings) {
```
XhmikosR, line 1: Just wondering, maybe `getActivePlugins` would be a better match?

XhmikosR:
This looks pretty good now! Waiting for @jakearchibald to review too.

KnifeFed:
No progress on the review? I guess https://deploy-preview-383--svgomg.netlify.app works for now though 🙂

AdrianoCahete:
JFYI: Using the preview build, I was able to export and use the config, but the property `cleanupIDs` returns an error:

```
Error: Unknown builtin plugin "cleanupIDs" specified.
    at resolvePluginConfig (...)
```

Mostly because `cleanupIDs` was renamed to `cleanupIds` in v3, as referenced in [changelog](https://github.com/svg/svgo/releases/tag/v3.0.0).

Manually renaming on config file works.

epodivilov:
> cleanupIds

Thanks for the feedback. Checked, there is indeed a problem. It is true that for SVGOMG to support the new plugin name you need to update the svgo package.

I updated my PR and added the necessary edits

PR #402, opened 2022-10-28
Disable `removeViewbox` plugin

XhmikosR:
While svgo still has this enabled by default, it can cause issues

See https://github.com/svg/svgo/pull/1461 for more details

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -103,7 +103,7 @@
     {
       "id": "removeViewBox",
       "name": "Remove viewBox",
-      "enabledByDefault": true
+      "enabledByDefault": false
     },
     {
       "id": "cleanupEnableBackground",


Darkproduct:
Oh, and don't forget to read this for a lot more detail: https://github.com/svg/svgo/issues/1128

Lucienest:
> While svgo still has this enabled by default, it can cause issues
> 
> See [svg/svgo#1461](https://github.com/svg/svgo/pull/1461) for more details

People were complaining about this for many years but the sole owner of this project won't listen.

coliff (approved):

SethFalco:
It will be some time before we make the change in SVGO (v4 release), so I'd recommend merging this into SVGOMG already.

It may also be worth adding either a warning, or including a link to the documentation for each plugin, so users are more informed on potential consequences of them.

https://svgo.dev/docs/plugins/removeViewbox/

jakearchibald:
My bigger plan here is to extend each plugin with its individual options

PowerKiKi:
If https://github.com/jakearchibald/svgomg/pull/457 is merged, then it will disable `removeViewBox` and `removeTitle` by default because this is the new default since svgo 4.0.

XhmikosR:
I also have [a branch](https://github.com/jakearchibald/svgomg/tree/xmr/dev) where I updated SVGO among other things and deployed this on Netlify https://incomparable-semifreddo-1417f3.netlify.app/

I would personally move forward with some of the PRs but ~~I don't have merge rights~~ I can't merge my own PRs without an approval, so only @jakearchibald can move things forward :)

PR #409, opened 2022-11-11
Update dependencies

XhmikosR:
Requires #400


src/css/components/_ripple.scss
```diff
@@ -1,12 +1,12 @@
 @keyframes ripple-pop {
-  from {
+  0% {
```
XhmikosR, line 2: @jakearchibald We can disable this stylelint rule if you prefer the old way (it's the same to me)

PR #413, opened 2022-11-13
fix: usage of await inside loop by using promises

akamfoad:
Instead of looping over in a for..of loop that uses await,
one can use Promise.all in which the promises are result
of filtering and mapping of the array using the Array.


jakearchibald (approved):
LGTM, thanks!

Issue #22, opened 2015-06-29
Easily Copy Out of Code View


shshaw:
I love using the Paste Markup & Code View to quickly optimize SVGs without having to mess with a lot of files. However, when I toggle over to Code View, I want to hit CMD + A to select all and copy the code out, but I always end up text from the UI:

```
Open SVG
Paste markup
Demo
Contribute
About
Powered by the incredible SVGO


IMAGE
CODE
765 bytes - 12.97% saving
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100"> ... </svg>
```

Any way to make it easier to copy the code view? Two suggestions:

* Make Code View a textarea or iframe so it's isolated from the rest of the document
* Add `user-select: none` to all the UI elements to prevent the text from being selected.

Tigt:
Adding the `contenteditable` attribute to the code's holder element seems like the simplest way of accomplishing this, and might also be nice to make a few edits the control panel can't.

shshaw:
Ah, good idea! Adding it to `.code-output` in DevTools gives the desired result.

jakearchibald:
Good idea, will play with this

s10wen:
:+1:

elwinschmitz:
:+1: 
Adding `spellcheck="false"`, would make the code output have less squiggly red lines.

And while we're at it: `autocomplete="off" autocorrect="off" autocapitalize="off"` would help if SVGOMG would support inline-editing of the SVG code...

PR #419, opened 2023-02-16
Easier copy/paste from the markup screen

gucr:
This pull request goal is to improve the user experience when copying the markup code in the markup screen.
As of today (and discussed in issue #22) some UI text gets copied along with the SVG markup when we use the keystrokes CMD+A —> CMD+C.

There is multiple ways to solve the issue:

- Excluding UI elements with `user-select:none`. Meaning that it will need to be specified if new UI regions appear in the future
- Excluding the whole app with `user-select:none`, and allowing some UI regions with with `user-select:auto`
- Making the output div `contentEditable="true"` but this requires to deactivate autocorrect, autocapitalize & so on. Also, this could lead to side effects whenever the contentEditable API evolves.
- For the future: using user-select:contain. This is CSS4 spec as we speak and [not yet supported by any browser](https://caniuse.com/mdn-css_properties_user-select_contain).

I went with the 1st solution (excluding `.main-menu` and `.action-button-container`) but I can adjust if needed

### DIFF
--- a/src/css/_main-layout.scss
+++ b/src/css/_main-layout.scss
@@ -45,6 +45,7 @@
   right: 13px;
   display: flex;
   align-items: center;
+  user-select: none;
 
   .minor-action-container {
     display: flex;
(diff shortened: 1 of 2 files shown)


XhmikosR:
It still isn't the proper way IMHO.

PR #450, opened 2024-09-09
Fix error when showing some special svg due to data url

### LINKED ISSUE #449: Cannot show some special svg
Reproduce steps:
- Open https://jakearchibald.github.io/svgomg/
- Open svg provided below
- We cannot see the svg

https://github.com/user-attachments/assets/397405d0-d002-4c0e-af72-d8ad9e0fe0e5

cesaryuan:
Close: #449

The bug https://bugzilla.mozilla.org/show_bug.cgi?id=1125667 in the code comment seems have been fixed.
image

So, it's time to use blob urls. It will resolve the bug #449.

### DIFF
--- a/src/js/page/ui/svg-output.js
+++ b/src/js/page/ui/svg-output.js
@@ -23,12 +23,10 @@ export default class SvgOutput {
   }
 
   setSvg({ text, width, height }) {
-    // TODO: revisit this
-    // I would rather use blob urls, but they don't work in Firefox
-    // All the internal refs break.
-    // https://bugzilla.mozilla.org/show_bug.cgi?id=1125667
     const nextLoad = this._nextLoadPromise();
-    this._svgFrame.src = `data:image/svg+xml,${encodeURIComponent(text)}`;
+    this._svgFrame.src = URL.createObjectURL(
+      new Blob([text], { type: 'image/svg+xml' }),
+    )
     this._svgFrame.style.width = `${width}px`;
     this._svgFrame.style.height = `${height}px`;
     return nextLoad;


jakearchibald:
Thanks for this! Does the SVG in https://github.com/jakearchibald/svgomg/issues/449 fail due to its size?

cesaryuan:
> Thanks for this! Does the SVG in https://github.com/jakearchibald/svgomg/issues/449 fail due to its size?

I have no idea for this😂

zunsthy:
> Thanks for this! Does the SVG in #449 fail due to its size?

no, #449 problem is because of no "xmlns" attribute. it's ok after toggle off the option "remove xmlns".

browser can't display normally "image/svg+xml" lack of "xmlns" attribute.

for example https://bugzilla.mozilla.org/show_bug.cgi?id=1125667 

it works
```xml
<svg xmlns="http://www.w3.org/2000/svg" width="100" height="100">
<defs><circle cx="50" cy="50" r="50" id="circle" /></defs>
<use href="#circle"/>
</svg>
```

it doesn't work
```xml
<svg width="100" height="100">
<defs><circle cx="50" cy="50" r="50" id="circle" /></defs>
<use href="#circle"/>
</svg>
```

i think need use 'text/html' as iframe source type, like this: 
```javascript
if (noXMLNS) {
   this._svgFrame.src = `data:text/html,${encodeURIComponent(`<html><head><style>body{margin:0}</style></head><body>${text}</body></html>`}`;
} else {
   this._svgFrame.src = URL.createObjectURL(new Blob([text], { type: 'image/svg+xml' }));
}
```

